PR #14, opened 2022-03-02
Updated geozero-shp crates to latest versions and Added a function to return the dbf FieldInfo.

apps4uco:
Updated geozero-shp crates to latest versions.

I made some modifications to get the crate to compile, hope they are ok.

Added a function to return the dbf FieldInfo. Useful for creating an sql table or a report

I discovered a problem accessing the FieldType in the dbf as it is in a private module.
so I created the following issue 

<https://github.com/tmontaigu/dbase-rs/issues/33#issue-1157794180>

The test shows how it could be used.

### DIFF
--- a/geozero-shp/src/property_processor.rs
+++ b/geozero-shp/src/property_processor.rs
@@ -7,29 +7,29 @@ impl FeatureProperties for ShapeRecord {
     /// Process feature properties.
     fn process_properties<P: PropertyProcessor>(&self, processor: &mut P) -> Result<bool> {
         let mut finish = false;
-        for (i, (name, value)) in self.record.iter().enumerate() {
+        for (i, (name, value)) in self.record.clone().into_iter().enumerate() {
             match value {
                 FieldValue::Character(Some(val)) => {
-                    finish = processor.property(i, name, &ColumnValue::String(val))?;
+                    finish = processor.property(i, &name, &ColumnValue::String(&val))?;
                 }
                 FieldValue::Numeric(Some(val)) => {
-                    finish = processor.property(i, name, &ColumnValue::Double(*val))?;
+                    finish = processor.property(i, &name, &ColumnValue::Double(val))?;
                 }
                 FieldValue::Logical(Some(val)) => {
-                    finish = processor.property(i, name, &ColumnValue::Bool(*val))?;
+                    finish = processor.property(i, &name, &ColumnValue::Bool(val))?;
                 }
                 FieldValue::Date(Some(_)) => {
                     let s = format!("{}", value);
-                    finish = processor.property(i, name, &ColumnValue::DateTime(&s))?;
+                    finish = processor.property(i, &name, &ColumnValue::DateTime(&s))?;
                 }
                 FieldValue::Float(Some(val)) => {
-                    finish = processor.property(i, name, &ColumnValue::Float(*val))?;
+                    finish = processor.property(i, &name, &ColumnValue::Float(val))?;
                 }
                 FieldValue::Integer(val) => {
-                    finish = processor.property(i, name, &ColumnValue::Int(*val))?;
+                    finish = processor.property(i, &name, &ColumnValue::Int(val))?;
                 }
                 FieldValue::Double(val) => {
-                    finish = processor.property(i, name, &ColumnValue::Double(*val))?;
+                    finish = processor.property(i, &name, &ColumnValue::Double(val))?;
                 }
                 _ => {}
             }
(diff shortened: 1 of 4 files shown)


geozero-shp/src/property_processor.rs
```diff
@@ -7,29 +7,29 @@ impl FeatureProperties for ShapeRecord {
     /// Process feature properties.
     fn process_properties<P: PropertyProcessor>(&self, processor: &mut P) -> Result<bool> {
         let mut finish = false;
-        for (i, (name, value)) in self.record.iter().enumerate() {
+        for (i, (name, value)) in self.record.clone().into_iter().enumerate() {
```
pka, line 10: Instead of `clone`, `as_ref` can be used.

geozero-shp/src/property_processor.rs
```diff
@@ -7,29 +7,29 @@ impl FeatureProperties for ShapeRecord {
...
             match value {
                 FieldValue::Character(Some(val)) => {
-                    finish = processor.property(i, name, &ColumnValue::String(val))?;
+                    finish = processor.property(i, &name, &ColumnValue::String(&val))?;
```
pka, line 13: When using `as_ref` this change can be reverted.

geozero-shp/src/property_processor.rs
```diff
@@ -7,29 +7,29 @@ impl FeatureProperties for ShapeRecord {
...
                 }
                 FieldValue::Numeric(Some(val)) => {
-                    finish = processor.property(i, name, &ColumnValue::Double(*val))?;
+                    finish = processor.property(i, &name, &ColumnValue::Double(val))?;
```
pka, line 16: When using `as_ref` this change can be reverted.

pka (changes requested):
Thanks a lot for giving the shape driver some love!

Could you please run `cargo fmt` and replace `clone` with `as_ref`?

geozero-shp/src/reader.rs
```diff
@@ -3,9 +3,10 @@ use crate::shx_reader::{read_index_file, ShapeIndex};
 use crate::{header, Error};
 use geozero::{FeatureProcessor, FeatureProperties, GeomProcessor};
 use std::fs::File;
-use std::io::{BufReader, Read};
+use std::io::{BufReader, Read, Seek};
 use std::iter::FusedIterator;
 use std::path::Path;
+pub use dbase::FieldInfo;
```
pka, line 9: `pub` is not strictly necessary here. Did you deliberately make it pub?

geozero-shp/src/property_processor.rs
```diff
@@ -7,29 +7,29 @@ impl FeatureProperties for ShapeRecord {
...
                 }
                 FieldValue::Integer(val) => {
-                    finish = processor.property(i, name, &ColumnValue::Int(*val))?;
+                    finish = processor.property(i, &name, &ColumnValue::Int(val))?;
                 }
                 FieldValue::Double(val) => {
-                    finish = processor.property(i, name, &ColumnValue::Double(*val))?;
+                    finish = processor.property(i, &name, &ColumnValue::Double(val))?;
                 }
                 _ => {}
```
pka, line 34: There seem to be new types which should be handled:
```
                FieldValue::Currency(val) => {
                    finish = processor.property(i, name, &ColumnValue::Double(*val))?;
                }
                FieldValue::DateTime(_) => {
                    let s = format!("{}", value);
                    finish = processor.property(i, name, &ColumnValue::DateTime(&s))?;
                }
                FieldValue::Memo(val) => {
                    finish = processor.property(i, name, &ColumnValue::String(val))?;
                }
                FieldValue::Character(None)
                | FieldValue::Numeric(None)
                | FieldValue::Logical(None)
                | FieldValue::Date(None)
                | FieldValue::Float(None) => {} // Ignore NULL values
```

geozero-shp/src/reader.rs
apps4uco: It was on purpose as I thought it would be good to reexport the FieldInfo and FieldType so that users would not need to declare the dbf crate just for those 2 types. 
And I wanted to match on FieldType in the example however I discovered that FIeldType is in a private module. Hence the issue I referred to.

geozero-shp/src/property_processor.rs
apps4uco: Ok thanks for the tip. I just pushed the changes.

This is my second PR so thanks for bearing with me, should I do anything else?

Issue #1, opened 2020-05-12
Implement a streaming GeoJSON reader


pka:
The current GeoJSON reader is based on https://github.com/georust/geojson which reads the whole file into memory.

A few unsuccessful attempts are in https://github.com/pka/geozero/tree/parse-geojson-with-serde:
- Serde Deserialize with custom deserialize methods for geometries and properties: I was not able to pass the parsed data out of the deserializon methods to the caller
- Serde Transcoder (https://serde.rs/transcode.html): too much semantic information is lost in the serialize methods. Enough for pretty printing, but not well suited for reading complex data structures
- Reading a GeoJSON stream and deserialize the current feature only: I didn't find a reliable way to detect the begin and end of a feature

I also had a deeper look at https://github.com/Geal/nom, which has a streaming interface. But I didn't find a good way to continue after a "Buffer empty" error.

frewsxcv:
a couple relevant issues in the popular json libraries:

- [`serde-rs/json#526`](https://github.com/serde-rs/json/pull/526)
- [`maciejhirsz/json-rust#143`](https://github.com/maciejhirsz/json-rust/issues/143)

pka:
Apache Arrow has a batch reader: https://github.com/apache/arrow/blob/master/rust/arrow/src/json/reader.rs

pka:
Rust library for encoding and decoding JSON streams.: https://docs.rs/destream_json/

pka:
https://github.com/jeremiah-shaulov/nop-json: Deserialization of JSON io::Read stream into primitive types, Strings and any other types.

pka:
https://github.com/serde-rs/json-benchmark/blob/master/src/canada.rs contains a GeoJSON benchmark.

bjornharrtell:
Arrow implementation is now at https://github.com/apache/arrow-rs/blob/master/arrow/src/json/reader.rs.

kylebarron:
Arrow implementation is now at https://github.com/apache/arrow-rs/blob/master/arrow-json/src/reader/mod.rs 🙃

kylebarron:
Was this issue intended to be closed by #16?

michaelkirk:
I think so!

PR #16, opened 2022-03-20
Add FeatureIterator for streaming deserialize

bjornharrtell:
~~This also includes the FeatureIterator from https://github.com/georust/geojson/pull/181 until further notice.~~
UPDATE: now uses it from released geojson 0.22.3

I'm unsure why the WKT in from_file_fc test differs from the from_file test in that WKT is not comma separated.

### DIFF
--- a/geozero/Cargo.toml
+++ b/geozero/Cargo.toml
@@ -28,7 +28,9 @@ with-tessellator = ["lyon"]
 [dependencies]
 thiserror = "1.0"
 geojson = { version = "0.22.3", optional = true }
+serde = "1.0.79"
 serde_json = "1.0.79"
+serde_derive = "1.0.79"
 geo-types = { version = "0.7", default-features = false, optional = true }
 geos = { version = "8.0", optional = true }
 gdal = { version = "0.12", default-features = false, optional = true }
--- a/geozero/Cargo.toml
+++ b/geozero/Cargo.toml
@@ -28,9 +28,7 @@ with-tessellator = ["lyon"]
 [dependencies]
 thiserror = "1.0"
 geojson = { version = "0.22.3", optional = true }
-serde = "1.0.79"
 serde_json = "1.0.79"
-serde_derive = "1.0.79"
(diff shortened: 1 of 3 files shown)


pka:
Looks like a good solution for solving #1, thanks!

Pushed some changes in branch `bjornharrtell-feature_iterator`. Maybe it's possible to push directly into this PR?

bjornharrtell:
@pka I've sent and invite to my fork, that should hopefully allow you to push the fixes directly to my branch.

pka:
When using `read_geojson_fc` as default function in `process` a few tests fail. Do we need a way to detect bare geometry GeoJSON objects of a fallback when FeatureIterator fails?

bjornharrtell:
@pka well I suppose so, read_geojson_fc assumes feature collection. I don't quite understand the api design with default function in process etc.

pka:
The `process` method is called by the conversion traits like  `GeoJson(r#"{"type": "Point", "coordinates": [10,20]}"#).to_wkt()`. So it has to support all GeoJSON object types.

bjornharrtell:
@pka I can perhaps see why that is needed technically right now doesn't quite sense to me logically that `read_geojson_fc` and `read_geojson_geom` need to be aware of other types than the name of the function implies.

pka:
Could you add 500c91f02bc8eca6379674ea4a45e61500db0323 again?

geozero/Cargo.toml
```diff
@@ -28,7 +28,9 @@ with-tessellator = ["lyon"]
 [dependencies]
 thiserror = "1.0"
 geojson = { version = "0.22.3", optional = true }
+serde = "1.0.79"
 serde_json = "1.0.79"
+serde_derive = "1.0.79"
```
pka, line 33: These dependencies aren't necessary anymore, I think?

bjornharrtell: Ah yes, removed.

bjornharrtell:
https://github.com/georust/geozero/commit/500c91f02bc8eca6379674ea4a45e61500db0323 is back in

PR #17, opened 2022-03-20
Upgrade deps

bjornharrtell:


geozero/Cargo.toml
```diff
@@ -27,28 +27,28 @@ with-tessellator = ["lyon"]
 [dependencies]
 thiserror = "1.0"
 geojson = { version = "0.22", optional = true }
-serde_json = "1.0.68"
+serde_json = "1.0"
```
michaelkirk, line 30: cargo dependency resolution is different from platforms.  As such it's usually preferable to leave the patch version in. 

https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277

This surprises a lot of people (myself included).

PR #18, opened 2022-03-25
fix issue processing certain types at start of ShapeRecord

jcary741:
I ran into a situation when reading a shapefile into geojson where, if the first property of a given record is null (and thereby omitted by the processor), a comma is written by the GeoJsonWriter for the next property resulting in invalid JSON like `{, "second_property": 123}`

This change offsets the `i` value given to the GeoJsonWriter (or whatever writer) when we're skipping null properties.

### DIFF
--- a/geozero-shp/src/property_processor.rs
+++ b/geozero-shp/src/property_processor.rs
@@ -6,53 +6,36 @@ use geozero::{ColumnValue, FeatureProperties, PropertyProcessor};
 impl FeatureProperties for ShapeRecord {
     /// Process feature properties.
     fn process_properties<P: PropertyProcessor>(&self, processor: &mut P) -> Result<bool> {
-        let mut finish = false;
-        let mut null_offset = 0usize;
-        for (i, (name, value)) in self.record.as_ref().iter().enumerate() {
-            let i = i - null_offset;
-            match value {
-                FieldValue::Character(Some(val)) => {
-                    finish = processor.property(i, name, &ColumnValue::String(val))?;
-                }
-                FieldValue::Numeric(Some(val)) => {
-                    finish = processor.property(i, name, &ColumnValue::Double(*val))?;
-                }
-                FieldValue::Logical(Some(val)) => {
-                    finish = processor.property(i, name, &ColumnValue::Bool(*val))?;
-                }
+        let mut i = 0;
+        for (name, value) in self.record.as_ref().iter() {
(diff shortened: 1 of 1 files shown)


pka:
Thanks for your PR!
What do you think about the more verbose variant:
```rust
        let mut i = 0;
        for (name, value) in self.record.as_ref().iter() {
            match value {
                FieldValue::Character(Some(val)) => {
                    finish = processor.property(i, name, &ColumnValue::String(val))?;
                    i += 1;
                }
                FieldValue::Numeric(Some(val)) => {
                    finish = processor.property(i, name, &ColumnValue::Double(*val))?;
                    i += 1;
                }
                FieldValue::Logical(Some(val)) => {
                    finish = processor.property(i, name, &ColumnValue::Bool(*val))?;
                    i += 1;
                }
                FieldValue::Date(Some(_)) => {
                    let s = format!("{}", value);
                    finish = processor.property(i, name, &ColumnValue::DateTime(&s))?;
                    i += 1;
                }
                FieldValue::Float(Some(val)) => {
                    finish = processor.property(i, name, &ColumnValue::Float(*val))?;
                    i += 1;
                }
                FieldValue::Integer(val) => {
                    finish = processor.property(i, name, &ColumnValue::Int(*val))?;
                    i += 1;
                }
                FieldValue::Double(val) => {
                    finish = processor.property(i, name, &ColumnValue::Double(*val))?;
                    i += 1;
                }
                FieldValue::Currency(val) => {
                    finish = processor.property(i, name, &ColumnValue::Double(*val))?;
                    i += 1;
                }
                FieldValue::DateTime(_) => {
                    let s = format!("{}", value);
                    finish = processor.property(i, name, &ColumnValue::DateTime(&s))?;
                    i += 1;
                }
                FieldValue::Memo(val) => {
                    finish = processor.property(i, name, &ColumnValue::String(val))?;
                    i += 1;
                }
                FieldValue::Character(None)
                | FieldValue::Numeric(None)
                | FieldValue::Logical(None)
                | FieldValue::Date(None)
                | FieldValue::Float(None) => {} // Ignore NULL values
            }
```

jcary741:
Ha! And here I was just trying to make enumerate work 😛

I took your suggestion one step further (see latest commit) by just assigning the result of the match statement, resulting in overall fewer lines of code.

pka:
Thanks!

PR #20, opened 2022-03-30
wkt reader

### LINKED ISSUE #19: implement WKT Reader
My first thought would be to build something on top of https://github.com/georust/wkt as you've done with the GeoJson reader. 

Have you put any thought into this already? I'm willing to work on this if you're interested @pka.

michaelkirk:
Fixes #19 

~~I have a couple things I was hoping to get input on before submitting for actual review... questions inline.~~
Update: OK! Ready for review!

Definitely some weirdness around handling `POINT EMPTY`, but I went with what felt to me like a reasonable compromise of trying to:

- handle `POINT EMPTY` where possible. Currently this is just Wkt->Wkt which isn't very likely to be used, but I suppose if someone wrote a custom processor, they might want to have the info.
- Ignore `EMPTY` in a MultiPoint.
- Else return `Err` when encountering a `POINT EMPTY`. I couldn't find a way to "be lazy" and ignore a `POINT EMPTY` for a top level geometry - we have to return *something*. And having `Wkt("POINT EMPTY).to_geo() -> point!(x: 0, y: 0)` seems likely to be a bad surprise.

### DIFF
--- a/geozero/src/wkt/wkt_reader.rs
+++ b/geozero/src/wkt/wkt_reader.rs
@@ -1,0 +1,343 @@
+use crate::error::{GeozeroError, Result};
+use crate::{FeatureProcessor, GeomProcessor, GeozeroDatasource, GeozeroGeometry};
+
+use std::io::Read;
+use wkt::types::{
+    Coord, GeometryCollection, LineString, MultiLineString, MultiPoint, MultiPolygon, Point,
+    Polygon,
+};
+use wkt::Geometry;
+
+/// WKT String.
+#[derive(Debug)]
+pub struct WktString(pub String);
+
+impl GeozeroGeometry for WktString {
+    fn process_geom<P: GeomProcessor>(&self, processor: &mut P) -> Result<()> {
+        read_wkt(&mut self.0.as_bytes(), processor)
+    }
+}
+
+/// WKT String slice.
+pub struct WktStr<'a>(pub &'a str);
+
+impl GeozeroGeometry for WktStr<'_> {
+    fn process_geom<P: GeomProcessor>(&self, processor: &mut P) -> Result<()> {
+        read_wkt(&mut self.0.as_bytes(), processor)
+    }
+}
+
+impl GeozeroDatasource for WktStr<'_> {
+    fn process<P: FeatureProcessor>(&mut self, processor: &mut P) -> Result<()> {
+        read_wkt(&mut self.0.as_bytes(), processor)
+    }
+}
+
+/// Wkt Reader.
+pub struct WktReader<'a, R: Read>(pub &'a mut R);
+
+impl<'a, R: Read> GeozeroDatasource for WktReader<'a, R> {
+    fn process<P: FeatureProcessor>(&mut self, processor: &mut P) -> Result<()> {
+        read_wkt(&mut self.0, processor)
+    }
+}
+
+/// Read and process WKT geometry.
+pub fn read_wkt<R: Read, P: GeomProcessor>(reader: &mut R, processor: &mut P) -> Result<()> {
+    use std::str::FromStr;
+    // PERF: it would be good to avoid copying data into this string when we already
+    // have a string as input. Maybe the wkt crate needs a from_reader implementation.
+    let mut wkt_string = String::new();
+    reader.read_to_string(&mut wkt_string)?;
+    let wkt = wkt::Wkt::from_str(&wkt_string).map_err(|e| GeozeroError::Geometry(e.to_string()))?;
+    process_wkt_geom(&wkt.item, processor)
+}
+
+/// Process WKT geometry
+fn process_wkt_geom<P: GeomProcessor>(geometry: &Geometry<f64>, processor: &mut P) -> Result<()> {
+    process_wkt_geom_n(geometry, 0, processor)
+}
+
+fn process_wkt_geom_n<P: GeomProcessor>(
+    geometry: &Geometry<f64>,
+    idx: usize,
+    processor: &mut P,
+) -> Result<()> {
+    let multi_dim = processor.multi_dim();
+    match geometry {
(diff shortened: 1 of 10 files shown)


geozero/src/wkt/wkt_reader.rs
```diff
@@ -0,0 +1,344 @@
...
+        read_wkt(&mut self.0, processor)
+    }
+}
+
+/// Read and process WKT geometry.
+pub fn read_wkt<R: Read, P: GeomProcessor>(reader: &mut R, processor: &mut P) -> Result<()> {
+    use std::str::FromStr;
+    // PERF: it would be good to avoid copying data into this string when we already
+    // have a string as input. Maybe the wkt crate needs a from_reader implementation.
+    let mut wkt_string = String::new();
```
michaelkirk, line 50: Most of this file was copied from the geojson_reader. Like there we copy the input into a String, which seems wasteful, but I don't know of another way around it.

I don't think the wkt crate currently exposes anything other than a from_str constructor.

Does this seem reasonable for now?

geozero/src/wkt/wkt_reader.rs
```diff
@@ -0,0 +1,344 @@
...
+    point: &Point<f64>,
+    multi_dim: bool,
+    idx: usize,
+    processor: &mut P,
+) -> Result<()> {
+    processor.point_begin(idx)?;
+    if let Some(ref coord) = point.0 {
+        process_coord(coord, multi_dim, 0, processor)?;
+    } else {
+        return Err(GeozeroError::Geometry(
```
michaelkirk, line 114: AFAIK WKT is the only format with an "empty point".

I think erroring here is the safest thing to do here given that there is no support for it in other formats, so it's not really defined how they should handle it.

Another option would be to add
```
pub trait GeomProcessor {
...
    fn empty_coordinate() -> Result {
         Err(GeozeroError::Geometry("The WKT Point was empty, but Points cannot be empty in other formats".to_string()))
    }
}
```

And then only override it in the WktReader/WktWriter, but I'm not sure if it would be useful for anything. WDYT @pka?

pka: Paul Ramsey from PostGIS calls ignoring empty geometries strategic laziness (https://gis.stackexchange.com/a/106942). In the meantime PostGIS supports empty geometries and so does GEOS AFAIK. I would continue with strategic laziness and choose the cheapest solution. What happens, if we emit no coordinate and don't return an error?

geozero/src/wkt/wkt_reader.rs
```diff
@@ -107,15 +107,14 @@ fn process_point<P: GeomProcessor>(
...
-    processor.point_begin(idx)?;
     if let Some(ref coord) = point.0 {
+        processor.point_begin(idx)?;
         process_coord(coord, multi_dim, 0, processor)?;
+        processor.point_end(idx)
     } else {
-        return Err(GeozeroError::Geometry(
-            "The WKT Point was empty, but Points cannot be empty in other formats".to_string(),
-        ));
+        // skip processing of POINT EMPTY for now since no other formats support it
```
pka: Is
```rust
    processor.point_begin(idx)?;
    if let Some(ref coord) = point.0 {
        process_coord(coord, multi_dim, 0, processor)?;
    } else {
        // POINT EMPTY
    }
    processor.point_end(idx)
```
giving problems? It looks like the correct representation of an empty geometry to me.

michaelkirk: I think the spelling is `POINT EMPTY` not `POINT ()`

geozero/src/wkt/wkt_reader.rs
michaelkirk: I tried being as lazy as possible with https://github.com/georust/geozero/pull/20/commits/dc6a63a15962970405623e02397cb0af6d2dd947

But I didn't like the results.
```
    #[test]
     fn empty_point() {
         let wkt = WktStr("POINT EMPTY");
         let actual = wkt.to_geo().unwrap();

         // This is weird - but it's not obvious to me what we should do in this case.
         let expected: geo_types::Geometry<f64> = point!(x: 0.0, y: 0.0).into();

         assert_eq!(expected, actual);
     }
```

We can't just "skip" an empty point - since the return type is not optional. It strikes me as pretty bad to return (0, 0). 

I've doubled back and reimplemented a "return error" approach when trying to process an empty point into anything other than that WKTWriter.

WDYT?

pka:
Looks like we can't avoid going into that rabbit hole. At least we're not alone (https://github.com/locationtech/jts/pull/567 https://trac.osgeo.org/geos/ticket/1005). Looks like your solution with an explicit `empty_point` is the cleanest one. But in the end we would probably need this for all geometry types?

Anyway let's merge this and improve handling of empty geometries seperately. Thanks a lot!

michaelkirk:
>  in the end we would probably need this for all geometry types?

Luckily, I think `POINTS EMPTY` is kind of an exception. It seems more reasonable to handle other empty geometries like `MULTIPOINT EMPTY`, `LINESTRING EMPTY`, etc. without exploding.

I've expounded on this here: https://github.com/georust/geozero/pull/21

geozero/src/wkt/wkt_reader.rs
phayes: What about representing empty points as ` geo_types::Geometry<f64> = point!(x: f64::NaN, y: f64::NaN).into();`  ?

urschrei: `geo` algorithms assume the absence of NaN values in order to produce correct results, and they _don't_ check for it (for performance reasons), so this would be a source of catastrophic downstream errors (what's the Euclidean distance between `Point(1.0)` and `Point(NaN, NaN)`) and / or a catastrophic perf regression (e.g. checking 10e7 points for NaN values)

PR #22, opened 2022-04-13
wip: proj transform

michaelkirk:
This is a demo of an API I'd like to be able to use with geozero to do more kinds of processing with the library.

Currently, it seems like processing is usually limited to "input -> output", which works great for converting formats and for some kinds of processing.

But often I want to do some kind of pipeline of operations. One such example here:

1. read from some format
2. project the points
3. write to some other format

```
let from = "EPSG:2230";
let to = "EPSG:26946";
let proj = Proj::new_known_crs(from, to, None).unwrap();

let wkt = WktStr("POINT (4760096.421921 3744293.729449)");
let expected = r#"{"type": "Point", "coordinates": [1450880.2910605022,1141263.0111604782]}"#;
assert_eq!(expected, &wkt.to_projected(proj).to_json().unwrap());
```

I personally think the user facing API is pretty good - but implementing it has some problems that I'll highlight below.


geozero/src/proj.rs
```diff
@@ -0,0 +1,139 @@
...
+        idx: usize,
+    ) -> Result<()> {
+        let (x, y) = self.proj.convert((x, y)).unwrap();
+        self.output.coordinate(x, y, z, m, t, tm, idx)
+    }
+
+    fn xy(&mut self, x: f64, y: f64, idx: usize) -> Result<()> {
+        let (x, y) = self.proj.convert((x, y)).unwrap();
+        self.output.xy(x, y, idx)
+    }
```
michaelkirk, line 99: This is the crux - we transform the points before passing them in to the "next" processor in the pipeline.

We only "need" these two methods `coordinate` / `xy`, but we have to explicitly delegate our other impls to the "output" processor, otherwise we'd inherit the "no-op" implementations for the GeomProcessor trait.

Effectively, I'm going for some kind of interceptor pattern here.

But it's very verbose. And also, it'd be easy to break if we added a new method to GeomProcessor/FeatureProcessor/PropertyProcessor and failed to update these long list of delegations below.

pka: Sorry, I'm currently short on time and leaving for Easter holidays soon. My only input would be that maybe a delegation pattern like https://github.com/flatgeobuf/flatgeobuf/blob/845111d2ba6765c8f43379877eb1f3440b78d4b6/src/rust/src/file_writer.rs#L302 could work here as well.

michaelkirk: > maybe a delegation pattern

If I understand correctly, that's exactly what I'm doing - I did use the `delegate` crate macro to make it a *little* more concise, but it seems like a *lot* of boilerplate.

I'm searching for, but have so far failed to find, a better way to do this.

pka:
I'm currently experimenting with a new event based API, which should support chaining of processors. Other advantages of an event stream would be serializing and buffering.

michaelkirk:
Excited to see what you come up with!

Issue #24, opened 2022-05-03
CSV reader/writer


michaelkirk:
Do you think a CSV reader would be worthwhile?

ogr2ogr supports it: https://gdal.org/drivers/vector/csv.html

For example I have a dataset like this:
```
Address,Type,Datetime,Latitude,Longitude,Report Location,Incident Number
904 7th Av,Car Fire,05/22/2019 12:55:00 PM,47.6069,-122.329051,POINT (-122.329051 47.6069),F190051945
9610 53rd Av S,Aid Response,05/22/2019 12:55:00 PM,47.515984,-122.266529,POINT (-122.266529 47.515984),F190051946
```

There are multiple ways that the csv might contain geometry data in the wild, some some thought about configuration is due. Two popular ones I've seen are:

1. If it's a table of points, lat and lon columns are specified.
2. wkt column

My real world example above actually has both!

pka:
I use the OGR CSV driver regularly, so yes, it would be worthwile.

PR #30, opened 2022-05-11
csv reader

michaelkirk:
The reader portion of #24. 

I've hit a bump while working on the writer portion, so that might need to wait a bit. Also, if I'm being honest, I only personally need the reader portion at the moment, so am being lazy. 😄

### DIFF
--- a/geozero-cli/Cargo.toml
+++ b/geozero-cli/Cargo.toml
@@ -10,7 +10,7 @@ name = "geozero"
 path = "src/main.rs"
 
 [dependencies]
-geozero = "0.9.0"
+geozero = { version = "0.9.4", features = ["with-csv"] }
 flatgeobuf = "0.8.0"
 async-trait = "0.1"
 clap = { version = "3.1.18", features = ["derive"] }
--- a/geozero/Cargo.toml
+++ b/geozero/Cargo.toml
@@ -12,6 +12,7 @@ keywords = ["geo", "geojson", "gdal", "geos", "postgis"]
 
 [features]
 default = ["with-svg", "with-wkt", "with-geo", "with-geojson"]
+with-csv = ["csv", "with-wkt"]
 with-svg = []
 with-wkt = ["wkt"]
 with-geo = ["geo-types"]
@@ -26,6 +27,7 @@ with-mvt = ["prost"]
 with-tessellator = ["lyon"]
 
 [dependencies]
+csv = { version = "1.1.6", optional = true }
 thiserror = "1.0"
 geojson = { version = "0.22.3", optional = true }
 serde_json = "1.0.79"
@@ -34,6 +36,7 @@ geos = { version = "8.0", optional = true }
 gdal = { version = "0.12", default-features = false, optional = true }
 gdal-sys = { version = "0.6", optional = true }
 lyon = { version = "0.17", optional = true }
+log = "0.4.17"
 scroll = { version = "0.11", optional = true }
 sqlx = { version = "0.5", default-features = false, optional = true }
 postgres-types = { version = "0.2", optional = true }
--- a/geozero/src/csv/csv_reader.rs
+++ b/geozero/src/csv/csv_reader.rs
@@ -1,0 +1,422 @@
+use crate::error::{GeozeroError, Result};
+use crate::{ColumnValue, FeatureProcessor, GeomProcessor, GeozeroDatasource, GeozeroGeometry};
+
+use std::io::Read;
+
+pub struct Csv<'a> {
+    csv_text: &'a str,
+    geometry_column_name: String,
+}
+
+impl<'a> Csv<'a> {
+    pub fn new(geometry_column_name: &str, csv_text: &'a str) -> Self {
+        Self {
+            csv_text,
+            geometry_column_name: geometry_column_name.to_string(),
+        }
+    }
+}
+
+impl GeozeroDatasource for Csv<'_> {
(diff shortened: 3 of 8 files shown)


geozero/Cargo.toml
```diff
@@ -34,6 +36,7 @@ geos = { version = "8.0", optional = true }
 gdal = { version = "0.12", default-features = false, optional = true }
 gdal-sys = { version = "0.6", optional = true }
 lyon = { version = "0.17", optional = true }
+log = "0.4.17"
```
michaelkirk, line 39: Note that I added the log crate. I'm only using it in one place at the moment, but a lot of our error messages are kind of opaque, and logging is an easy way to add context about line number, etc.

geozero/src/lib.rs
```diff
@@ -28,6 +28,9 @@
 //! | WKB       | [Wkb](wkb::Wkb), [Ewkb](wkb::Ewkb), [GpkgWkb](wkb::GpkgWkb)          | XYZM       | -                                                                    | [ToWkb]             | [WkbWriter](wkb::WkbWriter)             |
 //! | WKT       | [wkt::WktStr], [wkt::WktString]                                      | XYZM       | [wkt::WktReader], [wkt::WktStr], [wkt::WktString]                    | [ToWkt]             | [WktWriter](wkt::WktWriter)             |
 
+#[macro_use]
+extern crate log;
```
michaelkirk, line 32: I usually include the macros like this for log crate, so that I can `info!` instead of `log::info!`, but happy to revert it if it's controversial.

geozero-cli/Cargo.toml
```diff
@@ -10,7 +10,7 @@ name = "geozero"
 path = "src/main.rs"
 
 [dependencies]
-geozero = "0.9.0"
+geozero = "0.9.4"
```
michaelkirk, line 13: We'll need to bump this again when geozero gets released to be sure the CLI is using a version with CSV support. Alternatively I can bump the geozero/Cargo.toml version in this PR if you'd prefer.

geozero/Cargo.toml
```diff
@@ -11,7 +11,8 @@ license = "MIT/Apache-2.0"
 keywords = ["geo", "geojson", "gdal", "geos", "postgis"]
 
 [features]
-default = ["with-svg", "with-wkt", "with-geo", "with-geojson"]
+default = ["with-csv", "with-svg", "with-wkt", "with-geo", "with-geojson"]
```
pka, line 14: Inclusion as default feature disputable. The (also disputable) rule until now was: fhe feature either adds no dependency or the format is important enough for adding a light dependency.

geozero/src/csv/csv_reader.rs
```diff
@@ -0,0 +1,417 @@
...
+        let record = record?;
+        let geometry_field = record
+            .get(geometry_idx)
+            .ok_or(GeozeroError::ColumnNotFound)?;
+        use std::str::FromStr;
+        let wkt = wkt::Wkt::from_str(&geometry_field)
+            .map_err(|e| GeozeroError::Geometry(e.to_string()))?;
+        crate::wkt::wkt_reader::process_wkt_geom_n(&wkt.item, record_idx, processor).map_err(
+            |e| {
+                log::warn!("invalid WKT: {}", geometry_field);
```
pka, line 129: I would expect skipping this geometry error, when a warning is issued? Is this a debugging leftover?

pka (approved):
Thanks for this useful datasource implementation!

geozero/src/lib.rs
pka: I usually `use log::info`, but I'm also fine with this.

geozero/Cargo.toml
michaelkirk: I'm happy to have it non-default in the geozero library for now. Note that I've enabled the feature in the geozero-cli. 

Does that sound alright?

geozero/src/csv/csv_reader.rs
michaelkirk: It was intentional. I thought it made sense to fail when encountering invalid data, but I wanted to at least offer the user a way to track down the offending row.

Is there other precedence for how to handle partially readable data in the library? I'm a bit nervous about dropping data, potentially silently if the use hasn't enabled logging or maybe isn't even using the library in a text/terminal context.

michaelkirk: I just pushed an update with a better error message - matching the similar method in the FeatureProcessor impl below.

geozero/src/lib.rs
michaelkirk: I've removed this in favor of `log::warn!`, I don't have strong feelings about it either and now that csv is optional it was producing an "unsused" error. 🙃

michaelkirk:
Ok, I believe I've responded to all your feedback @pka.

The last remaining thing is if you're OK to fail hard when encountering an invalid WKT. 

That's my preference, but I'd be open to some kind of `--strict`/`--no-strict` configuration that gets passed into the reader.

pka:
Thanks!

One point was maybe a misunderstanding. I include the macro with `use log::warn` and then also log with `warn!("...")`. But that's nitpicking.

I'm fine with both hard failing on a WKT error or skipping with warning. Was just not sure whether it was intentional. I let you decide.

Could you please squash the CSV commits for merging?

michaelkirk:
Squashed!

PR #33, opened 2022-05-13
csv writer

### LINKED ISSUE #24: CSV reader/writer

michaelkirk:
FIXES #24 

This is a follow up to #30, but for writing CSV's.

Feel free to ask questions or push back on the implementation - I'm not super happy with it.

### DIFF
--- a/geozero/src/csv/csv_writer.rs
+++ b/geozero/src/csv/csv_writer.rs
@@ -1,0 +1,408 @@
+use crate::error::Result;
+use crate::{ColumnValue, CoordDimensions, FeatureProcessor, GeomProcessor, PropertyProcessor};
+use buffering_wkt_writer::BufferingWktWriter;
+
+use std::io::Write;
+
+pub struct CsvWriter<'w, W: Write> {
+    csv: csv::Writer<&'w mut W>,
+    headers: Vec<String>,
+    has_written_first_record: bool,
+    current_row_props: Vec<String>,
+    wkt_writer: BufferingWktWriter,
+}
+
+impl<'w, W: Write> CsvWriter<'w, W> {
+    pub fn new(out: &'w mut W) -> Self {
+        let wkt_writer = BufferingWktWriter::new();
+        Self {
+            csv: csv::Writer::from_writer(out),
+            headers: vec!["geometry".to_string()],
+            has_written_first_record: false,
+            current_row_props: vec![],
+            wkt_writer,
+        }
+    }
+
+    pub fn set_dims(&mut self, dims: CoordDimensions) {
+        self.wkt_writer.dims = dims;
+    }
+}
+
+impl<W: Write> FeatureProcessor for CsvWriter<'_, W> {
+    fn dataset_begin(&mut self, _name: Option<&str>) -> Result<()> {
+        debug_assert_eq!(self.headers, &["geometry"]);
+        Ok(())
+    }
+    fn dataset_end(&mut self) -> Result<()> {
+        self.headers = vec!["geometry".to_string()];
+        Ok(())
+    }
+    fn feature_begin(&mut self, _idx: u64) -> Result<()> {
+        debug_assert!(self.current_row_props.is_empty());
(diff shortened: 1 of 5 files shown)


geozero/src/csv/csv_writer.rs
```diff
@@ -0,0 +1,408 @@
...
+        self.wkt_writer.tin_begin(size, idx)
+    }
+    fn tin_end(&mut self, idx: usize) -> Result<()> {
+        self.wkt_writer.tin_end(idx)
+    }
+}
+
+// Writing Wkt is delegated to the WktWriter, and it's output buffered to Vec, to be eventually
+// incorporated in the output of the CSV writer. Is there a better way?
+mod buffering_wkt_writer {
```
michaelkirk, line 215: And another entire level of delegation!!! 😱

geozero/src/csv/csv_writer.rs
```diff
@@ -0,0 +1,408 @@
...
+
+        // TODO: support non-string colval
+        self.current_row_props.push(colval.to_string());
+        Ok(false)
+    }
+}
+
+impl<W: Write> GeomProcessor for CsvWriter<'_, W> {
+    fn dimensions(&self) -> CoordDimensions {
+        self.wkt_writer.dimensions()
```
michaelkirk, line 101: All this explicit delegation is kind of tedious. But I'm not sure of a better way.

geozero/src/csv/csv_writer.rs
```diff
@@ -0,0 +1,408 @@
...
+                       "address": "9610 53rd Av S",
+                       "type": "Aid Response",
+                       "datetime": "05/22/2019 12:55:00 PM",
+                       "incident number": "F190051946"
+                   }
+               }
+            ]
+        });
+
+        let expected_output = r#"geometry,address,datetime,incident number,type
```
michaelkirk, line 397: note the column order here is different from what's listed in the geojson. By default geojson doesn't keep it's properties ordered. The underlying serde_json crate does offers this as an optional feature.  I opened an issue to discuss exposing it in geojson here: https://github.com/georust/geojson/issues/187

I'm not totally sure we'd want to use it, but it might be a nice option since we have a concept of "property idx" in some data formats.

geozero/src/geojson/geojson_reader.rs
```diff
@@ -96,7 +96,7 @@ fn process_geojson<P: FeatureProcessor>(gj: &GeoGeoJson, processor: &mut P) -> R
                 }
                 if let Some(ref geometry) = feature.geometry {
                     processor.geometry_begin()?;
-                    process_geojson_geom_n(geometry, idx, processor)?;
+                    process_geojson_geom_n(geometry, 0, processor)?;
```
michaelkirk, line 99: Note this change - It might be controversial!

Without it, the second line of the csv will have it's geometry prefixed with an unwanted ",".

I'm willing to be talked out of this if you have a good rationale.

Alternatively, I could revert this and hack around this in the CSVWriter, but I think it's arguably correct for the first geometry in each feature to have index 0.

michaelkirk:
Whoops - I'm going to close this. The test failures are indicative of another issue that I've been tripping on.

PR #34, opened 2022-05-13
FIX: feature_collection.to_geo() returns only last feature geometry

michaelkirk:
Now we return a geometry collection when the input has more than one geometry (e.g. a FeatureCollection)

### DIFF
--- a/geozero/src/geo_types/geo_types_writer.rs
+++ b/geozero/src/geo_types/geo_types_writer.rs
@@ -5,7 +5,7 @@ use std::mem;
 
 /// Generator for geo-types geometry type.
 pub struct GeoWriter {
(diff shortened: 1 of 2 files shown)


geozero/src/geo_types/geo_types_writer.rs
```diff
@@ -19,16 +19,23 @@ pub struct GeoWriter {
...
             coords: None,
             line_strings: None,
             polygons: None,
             collections: Vec::new(),
         }
     }
 
-    pub fn geometry(&self) -> Option<&Geometry<f64>> {
-        self.geom.as_ref()
+    pub fn take_geometry(&mut self) -> Option<Geometry<f64>> {
```
michaelkirk, line 30: Note this is a breaking change.

PR #36, opened 2022-05-13
csv writer

michaelkirk:
Ok, let's try this again (supercedes #33). 

FIXES https://github.com/georust/geozero/issues/24

This is a follow up to https://github.com/georust/geozero/pull/30, but for writing CSV's.

### DIFF
--- a/geozero/src/csv/csv_writer.rs
+++ b/geozero/src/csv/csv_writer.rs
@@ -1,0 +1,484 @@
+use crate::error::Result;
+use crate::{ColumnValue, CoordDimensions, FeatureProcessor, GeomProcessor, PropertyProcessor};
+use buffering_wkt_writer::BufferingWktWriter;
+
+use std::io::Write;
+
+pub struct CsvWriter<'w, W: Write> {
+    csv: csv::Writer<&'w mut W>,
+    headers: Vec<String>,
+    has_written_first_record: bool,
+    current_row_props: Vec<String>,
+    wkt_writer: BufferingWktWriter,
+}
+
+impl<'w, W: Write> CsvWriter<'w, W> {
+    pub fn new(out: &'w mut W) -> Self {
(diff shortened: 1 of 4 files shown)


geozero/src/csv/csv_writer.rs
```diff
@@ -0,0 +1,484 @@
...
+                       "address": "9610 53rd Av S",
+                       "type": "Aid Response",
+                       "datetime": "05/22/2019 12:55:00 PM",
+                       "incident number": "F190051946"
+                   }
+               }
+            ]
+        });
+
+        let expected_output = r#"geometry,address,datetime,incident number,type
```
michaelkirk, line 415: note the column order here is different from what's listed in the geojson. By default geojson doesn't keep it's properties ordered. The underlying serde_json crate does offers this as an optional feature. I opened an issue to discuss exposing it in geojson here: https://github.com/georust/geojson/issues/187

I'm not totally sure we'd want to use it, but it might be a nice option since we have a concept of "property idx" in some data formats.

geozero/src/csv/csv_writer.rs
```diff
@@ -0,0 +1,484 @@
...
+        pub(crate) fn bytes(&self) -> &[u8] {
+            &self.buffer
+        }
+
+        fn wkt_writer(&mut self) -> WktWriter<'_, Vec<u8>> {
+            WktWriter::new(&mut self.buffer)
+        }
+    }
+
+    impl GeomProcessor for BufferingWktWriter {
```
michaelkirk, line 260: All this explicit delegation is kind of tedious. But I'm not sure of a better way to embed the output of WktWriter into the CSVWriter.

pka: I hope these delegations won't be necessary with the event API anymore.

Issue #37, opened 2022-05-21
Geoparquet format reader


roger120981:
It is possible to incorporate in your library the new format launched by the OGC, GeoParquet..This is the link https://github.com/opengeospatial/geoparquet

pka:
A GeoPartquet reader would be welcome for sure. It shouldn't be very difficult to implement, since there are Rust implementations for reading Parquet (https://github.com/apache/arrow-rs) and AFAIK geometries are stored as WKB, which is already supported in geozero.

BKDaugherty:
I'm new to geospatial, but I'm gonna take a look at trying this out!

kylebarron:
Just a quick note that in versions of geoparquet so far WKB is the standard, but in future versions we hope to also enable an Arrow-native geometry encoding, see https://github.com/geopandas/geo-arrow-spec/blob/main/format.md. The difference between the two is that a binary buffer of WKB in memory still needs to be _parsed_ to be used. In contrast, an Arrow-native implementation would store geometries in a layout standardized by the Arrow project. This means that it's `O(1)` to access any individual coordinate.

What this means is that using geozero on a GeoParquet dataset with geometries stored as WKB is essentially already supported by `geozero`, you just iterate over the items in the array and use the existing geozero WKB reader. Note here that an improvement would be to figure out how to let the WKB reader read _directly from the Arrow buffer_. I seemed to have to [copy the section of the arrow buffer into a `Vec<u8>`](https://github.com/kylebarron/geopolars/blob/05826267b004e9948cb4ac6bc6341e2ea891c82f/geopolars/src/util.rs#L12-L13) to get geozero's WKB reader to accept it.

A separate `geozero-geoarrow` implementation would be to add support for the Arrow-native encoding. Note that https://github.com/geopandas/geo-arrow-spec/blob/main/format.md is a provisional, unreleased spec (though it was implemented by GDAL in 3.5.0). So if we have feedback while implementing a Rust reader, we should comment on the project. For prior work, there are WIP native arrow implementations in [Python](https://github.com/jorisvandenbossche/python-geoarrow), [C++](https://github.com/paleolimbot/geoarrow-cpp), [R](https://github.com/paleolimbot/geoarrow), and a [data repo](https://github.com/paleolimbot/geoarrow-data).

apps4uco:
I am starting work on a [geoparquet](https://crates.io/crates/geoparquet) crate that will use geozero and could potentially be merged into geozero at some point.

[...]

PR #39, opened 2022-06-02
GeoArrow WKB reader

kylebarron:
For https://github.com/georust/geozero/issues/37. As a disclaimer, this is my first Rust PR to a non-personal project, so suggestions welcomed 🙂. I'm new to the geozero API as well, so I tried to model after existing code. I figured I'd put up a draft PR to start some discussion.

Most likely [GeoArrow](https://github.com/geopandas/geo-arrow-spec)/[GeoParquet](https://github.com/opengeospatial/geoparquet) will end up having two geometry formats: WKB for universal compatibility and a "native" Arrow encoding for zero-copy performance. (For context, GeoArrow is the memory format, GeoParquet is the file format; GeoParquet decodes into GeoArrow). Is I mentioned in #37, the arrow-native encoding is faster because it's constant time access to any individual coordinate. (Though note that the arrow-native encoding is still provisional).

So it seems like there are four parts:

- GeoArrow WKB reader
- GeoArrow WKB writer
- GeoArrow Arrow-native reader
- GeoArrow Arrow-native writer

The WKB reader seems to be the easiest, so this PR starts there.

Open questions:

- I'm not sure the right way to handle a vector of geometries. I suppose the current use of `geometrycollection_begin` and `geometrycollection_end` exposes the collection of geometries as a `geo` `GeometryCollection`?

  Or should this implement `GeozeroDatasource` instead? The issue with that is that that seems to need to implement a _feature_ processor. And it's unclear whether the struct should require a single array or a chunk with multiple arrays. Should there be different structs for just the geometry array and for the entire table?
- Since the struct is defined as 

  ```rs
  pub struct GeoArrowWkb(pub BinaryArray<i32>);
  ```

  this takes ownership of the arrow array right? Seems like it would be preferable to take a reference here, but I don't really understand lifetimes yet 😅 .
- There are two different rust Arrow implementations: [`arrow`](https://github.com/apache/arrow-rs) and [`arrow2`](https://github.com/jorgecarleitao/arrow2). They're both pretty feature complete, so ideally `geozero` wouldn't enforce one or the other. I'm not sure the best way to do this... feature flags?


pka:
Wow, this was quick! Could you add some test data? Then I can play around and maybe answer your questions with code.

kylebarron:
> Could you add some test data?

I've been doing simple tests with a [point dataset of capital cities](https://github.com/kylebarron/geopolars/blob/05826267b004e9948cb4ac6bc6341e2ea891c82f/geopolars/cities.arrow), which is the result of
```py
import geopandas as gpd
gdf = gpd.read_file(gpd.datasets.get_path('naturalearth_cities'))
gdf.to_feather('cities.arrow', compression='uncompressed')
```

Alternatively, you should be able to convert any file to a WKB-encoded GeoArrow file with GDAL 3.5.0. Note that it defaults to the provisional arrow-native encoding, so you need to force output as WKB:
```bash
docker run --rm -v $(pwd):/data osgeo/gdal:latest \
  ogr2ogr \
  /data/output.arrow \
  /data/input.shp \
  -lco COMPRESSION=NONE \
  -lco GEOMETRY_ENCODING=WKB
```

And then you can use the `arrow2` IPC [`FileReader`](https://docs.rs/arrow2/latest/arrow2/io/ipc/read/struct.FileReader.html) to iterate over the Arrow record batches in the file.

michaelkirk:
It'd be great if you could include some small sample data and parsing tests like that directly in your PR. e.g. we have some sample data for the other formats here: https://github.com/georust/geozero/tree/master/geozero/tests/data

kylebarron:
```diff
-impl GeozeroGeometry for GeoArrowWkb {
+impl GeozeroGeometry for BinaryArray<i32> {
```
Interesting! So we don't need to define our own struct at all? We can just implement our traits for the arrow types directly.

Along the same lines, seems like there should be a way then to implement traits on both `arrow2`'s and `arrow`'s `BinaryArray`. Then it seems it would work out of the box for whichever package the user is already using.

pka:
> Interesting! So we don't need to define our own struct at all? We can just implement our traits for the arrow types directly.

Right, we can directly implement the trait for the arrow types.

> 
> Along the same lines, seems like there should be a way then to implement traits on both `arrow2`'s and `arrow`'s `BinaryArray`. Then it seems it would work out of the box for whichever package the user is already using.

Maybe we have to repeat the code for the different types or write a macro it there are too many similar implementations.

pka:
Found a generic way:

```rust
impl GeozeroGeometry for BinaryArray<i64> {
    fn process_geom<P: GeomProcessor>(&self, processor: &mut P) -> Result<()> {
        process_geoarrow_wkb_geom(&self, processor)
    }
}

fn process_geoarrow_wkb_geom<T: Offset>(
    array: &BinaryArray<T>,
    processor: &mut impl GeomProcessor,
) -> Result<()> {
  // ...
}
```

kylebarron:
I added a `countries.arrow` test data file (with geometries in WKB) generated from the [above command](https://github.com/georust/geozero/pull/39#issuecomment-1145109400). I made the suggested changes and put the `impl` directly on `BinaryArray` for `i32` and `i64`.

In the future, at least once https://github.com/geopandas/geo-arrow-spec/pull/22 is merged, we could also put an `impl` on the arrow _`ExtensionType`_, so that the conversion also works on arrays of data type `ExtensionType("geoarrow.wkb", BinaryArray<T>)`.

Let me know if you'd like more tests than the one @pka suggested above.

PR #42, opened 2022-06-24
update geozero internal usage

michaelkirk:


geozero-bench/Cargo.toml
```diff
@@ -9,7 +9,7 @@ repository = "https://github.com/georust/geozero"
 license = "MIT/Apache-2.0"
 
 [dependencies]
-geozero = { version = "0.7.0", features = ["with-geo", "with-geojson", "with-geos", "with-gpkg", "with-postgis-sqlx", "with-postgis-postgres"] }
+geozero = { path = "../geozero", features = ["with-geo", "with-geojson", "with-geos", "with-gpkg", "with-postgis-sqlx", "with-postgis-postgres"] }
```
michaelkirk, line 12: I *think* that we never publish this bench crate, so it should be fine (and more convenient) to just reference the local development crate.

pka: Right. But according to the test failures, the bench create needs some compatibiliy updates.

michaelkirk: oh sheesh. Sorry - I've pushed up another commit that's passing.

PR #43, opened 2022-06-29
FIX: spurious comma on subsequent rows

michaelkirk:
I encountered this same problem with my initial approach at a CSV writer (https://github.com/georust/geozero/pull/33), and incorporated a work-around in what we ultimately merged (https://github.com/georust/geozero/pull/36), but I failed to consistently apply the work-around, so it was still broken for some geometry types.

### DIFF
--- a/geozero/src/csv/csv_writer.rs
+++ b/geozero/src/csv/csv_writer.rs
@@ -135,92 +135,92 @@ impl<W: Write> GeomProcessor for CsvWriter<'_, W> {
         self.wkt_writer.point_end(self.offset_geom_idx(idx))
     }
     fn empty_point(&mut self, idx: usize) -> Result<()> {
-        self.wkt_writer.empty_point(idx)
+        self.wkt_writer.empty_point(self.offset_geom_idx(idx))
     }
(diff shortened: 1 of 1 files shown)


geozero/src/csv/csv_writer.rs
```diff
@@ -477,8 +477,29 @@ POINT(1 45),904 7th Av,05/22/2019 12:55:00 PM,F190051945,Car Fire
...
+            "type": "FeatureCollection",
+            "features": [
+                { "type": "Feature", "properties": { "NAME": "Foo" }, "geometry": { "type": "Polygon", "coordinates": [ [ [ 1, 1], [2, 2], [1, 1] ] ] } },
+                { "type": "Feature", "properties": { "NAME": "Bar" }, "geometry": { "type": "Polygon", "coordinates": [ [ [ 3, 1], [3, 2], [3, 1] ] ] } }
+            ]
+        }"#;
+
+        let expected_output = r#"geometry,NAME
+"POLYGON((1 1,2 2,1 1))",Foo
+"POLYGON((3 1,3 2,3 1))",Bar
```
michaelkirk, line 495: Without the fix, there's a spurious comma at the beggining of this second row: `,POLYGON((3 1...`

PR #45, opened 2022-08-24
Use `prost-build` only with `with-mvt` flag in build script

landreussi:
This PR makes the lib use `prost-build` only when the feature
`with-mvt` is enabled, once protobuf stuff is used only in this
feature.

I am opening this PR because I am working on a project that doesn't use
the `with-mvt` flag and protobufs and it spends 31s compiling this dep
 anyway.

I forked this repository, applied this patch, and used it in my project
and it worked flawlessly.
 
 I've run the tests and this patch doesn't break anything, but if there is something
 weird in this PR, please let me know :)


pka:
Great addition, thank you!

Issue #47, opened 2022-09-03
Add a GPX reader


frewsxcv:
https://github.com/georust/gpx

frewsxcv:
I'm looking into this

PR #48, opened 2022-09-04
GPX read support

frewsxcv:
Fixes https://github.com/georust/geozero/issues/47

This adds a `GeomProcessor` for GPX files


geozero/src/lib.rs
```diff
@@ -25,6 +25,7 @@
 //! | GeoJSON   | [GeoJson](geojson::GeoJson), [GeoJsonString](geojson::GeoJsonString) | XYZ        | [GeoJsonReader](geojson::GeoJsonReader), [GeoJson](geojson::GeoJson) | [ToJson]            | [GeoJsonWriter](geojson::GeoJsonWriter) |
 //! | GDAL      | `gdal::vector::Geometry`                                             | XYZ        | -                                                                    | [ToGdal]            | [GdalWriter](gdal::GdalWriter)          |
 //! | GEOS      | `geos::Geometry`                                                     | XYZ        | -                                                                    | [ToGeos]            | [GeosWriter](geos::GeosWriter)          |
+//! | GPX       |                                                                      | XY         |                                                                      |                     |                                         |
```
frewsxcv, line 28: What goes here?

geozero/src/gpx/gpx_reader.rs
```diff
@@ -0,0 +1,216 @@
...
+
+fn process_top_level_waypoints<P: crate::GeomProcessor>(
+    gpx_reader: &gpx::Gpx,
+    processor: &mut P,
+) -> crate::error::Result<()> {
+    if gpx_reader.waypoints.is_empty() {
+        return Ok(());
+    }
+    processor.multipoint_begin(gpx_reader.waypoints.len(), 0)?;
+    process_waypoints_iter(gpx_reader.waypoints.iter(), processor, 0)?;
```
frewsxcv, line 28: Should this `index` always be `0`?

geozero/src/lib.rs
pka: Would be the structs implementing `GeozeroGeometry`,  `GeomProcessor`, etc.

geozero/src/gpx/gpx_reader.rs
pka: `index` is the nth item within a parent item:
```md
                          +-----------------+
                          |                 |
                          |     Initial     <-------------+
                          |                 |             |
                          +--------^--------+    +--------v---------+
                                   |             |                  |
                             +-+-+---+-+         |GeometryCollection|
                             | | | | | |         |                  |
                             v v v v v v         +--------^---------+
+-----------------+                                       |
|                 |                                 +-+-+---+-+
|  MultiPolygon   |                                 | | | | | |
|                 |                                 v v v v v v
+--------^--------+   +-----------------+
         |            |                 |
         |            | MultiLineString |
+--------v--------+   |                 |
|                 |   +--------^--------+
|     Polygon     |            |
|                 |            |
+--------^--------+   +--------v--------+                         +-------------+   +--------------+
         |            |                 |                         |             |   |              |
         +----------->+    LineString   |                         |    Point    |   |  MultiPoint  |
                      |                 |                         |             |   |              |
                      +--------+--------+   +-----------------+   +------+------+   +-------+------+
                               |            |                 |          |                  |
                               +------------>   Coordinate    <----------+------------------+
                                            |                 |
                                            +-----------------+
```
E.g. the nth `LineString` within a `MultiLineString` or the nth `Coordinate` within a `LineString`

pka (changes requested):

geozero/src/gpx/gpx_reader.rs
frewsxcv: @pka Thanks for the explanation. Does the latest commit look correct? https://github.com/georust/geozero/pull/48/commits/80e82e33b019096e0cfb8b8ef62156b2886eaeeb

geozero/src/lib.rs
frewsxcv: I didn't add a new struct in this pull request. Should I?

pka: You don't have to, but if you add a `GpxReader` struct implementing `GeozeroDatasource` (like e.g. `GeoJsonReader`), users can do e.g.
`let wkt = GpxReader(&mut f).to_wkt()?;`

pka (approved):

geozero/src/gpx/gpx_reader.rs
```diff
@@ -0,0 +1,227 @@
...
+}
+
+fn process_top_level_routes<P: crate::GeomProcessor>(
+    gpx_reader: &gpx::Gpx,
+    processor: &mut P,
+) -> crate::error::Result<()> {
+    if gpx_reader.routes.is_empty() {
+        return Ok(());
+    }
+    processor.multilinestring_begin(gpx_reader.routes.len(), 0)?;
```
michaelkirk, line 95: Is it OK to always pass 0 here? IIRC CSV and maybe some other formats use the index to determine when to have a separating comma.

pka: Looking at `test_extensive` the output is a `GeometryCollection` with one `MultiPoint` (index 0) and two `MultiLineString` geometries (index 1+2).

michaelkirk: The tests are using the `geo-types` writer, which doesn't use the index parameter, so you could give it any value you wanted. The parameter exists because other writers/processors do use this parameter.

michaelkirk: I guess let me see if I can break it. =)

pka: As a comparison: `ogrinfo` shows 5 layers with the following geometry types:
- 1: waypoints (Point)
- 2: routes (Line String)
- 3: tracks (Multi Line String)
- 4: route_points (Point)
- 5: track_points (Point)

As we don't have multiple layers, we can have a collection containing the 3 first layers. I would not issue a MultiPoint for waypoints, since these points don't belong to a single feature.

frewsxcv: I introduced a `TestWriter` in my latest commit that better visualizes all the processor commands that are run: https://github.com/georust/geozero/pull/48/commits/c50a32f2629ba974bf1a293e37f85d43a34721b5

Regarding the GPX changes you both are talking about, I've lost the thread a little. What should I be changing? It sounds like waypoints should not be a `MultiPoint`.

michaelkirk: > I guess let me see if I can break it.

See #51 for some executable examples of issues I see with the current implementation.

Feel free to build off of it or just use it as a reference, and let me know if you think any of those tests are invalid.

frewsxcv:
@pka @michaelkirk This is ready for another review

pka:
The outputs in the tests look ok to me. Could you please "fix" them and reduce the git commits?

frewsxcv:
> The outputs in the tests look ok to me. Could you please "fix" them and reduce the git commits?

Done

frewsxcv:
> reduce the git commits?

Can we just use the "Squash and merge" feature in the GitHub pull request?

pka:
> > reduce the git commits?
> 
> Can we just use the "Squash and merge" feature in the GitHub pull request?

I'm afraid that @michaelkirk's commits are lost with "Squash and merge", so a manual `git rebase -i` session is needed.

michaelkirk:
FWIW - I'm happy to see this get merged with or without my commits staying in tact. Go wild!

frewsxcv:
https://github.blog/changelog/2019-12-19-improved-attribution-when-squashing-commits/

> Before today, whoever opened the pull request became the sole author of the squash commit. Now, we will automatically credit every commit author in the pull request as a co-author on the squash commit.

GitHub will attribute everyone in the PR

PR #53, opened 2022-10-02
ci: setup dependabot

sunng87:
This library has a few dependencies. Let's setup dependabot to keep them up-to-date.


sunng87:
Since this is relatively simple I'm going to merge it. Just let me know if your have any question. 

I'm adding this because we need a newer version of `gdal-sys` to work with latest `gdal`

PR #71, opened 2022-10-30
Sketch of reading from GeoArrow struct encoding

kylebarron:
This is a sketch (only point and multipoint are implemented for now) of reading from the [proposed struct encoding](https://github.com/geoarrow/geoarrow/pull/26) in the geoarrow spec.

GeoArrow is undecided as of now whether the point object should be represented by a logical "struct" (physically, the x and y arrays are distinct in memory) or a logical "fixed size list" (physically, x and y coordinates are interleaved). This PR implements the struct approach; it's not hard to switch between the two, so we could implement both for now and then remove one if/when the geoarrow spec chooses one for the standard.

### Notes:

- If feedback is good, I'd work to add the rest of the types. It's pretty straightforward, though ensuring offsets are all correct can be a little tedious (and will need tests ofc)
- Arrow extension types are used to clarify the meaning of various data types. For example a `List<Struct<x: f64, y: f64>>` (i.e. a logical type representing a variable size list of x,y structs) could either represent a multipoint or a linestring, so we need some metadata to describe which it is. Arrow2 has an [example here of using extension types](https://github.com/jorgecarleitao/arrow2/blob/f32b22e76e3af8265139d74df83cde6ce373c7d7/examples/extension.rs). Geoarrow extension type names are [defined here](https://github.com/geoarrow/geoarrow/blob/f2eb98e7e2f60ed409ed88078015a9c45f469604/extension-types.md).
- Not sure how to represent null geometries through geozero. In arrow, every value has a null bitmap, so every geometry can be null
- In the long term, I still want to work towards geo traits, but this seems a worthwhile stepping stone.

Related to https://github.com/georust/geozero/issues/37

### DIFF
--- a/geozero/src/arrow/geoarrow_struct_reader.rs
+++ b/geozero/src/arrow/geoarrow_struct_reader.rs
@@ -1,0 +1,149 @@
+use crate::error::Result;
+use crate::{GeomProcessor, GeozeroGeometry};
+use arrow2::array::{Array, ListArray, Offset, PrimitiveArray, StructArray};
+use arrow2::datatypes::DataType;
+
+/// A bare StructArray is a collection of points
+impl GeozeroGeometry for StructArray {
+    fn process_geom<P: GeomProcessor>(&self, processor: &mut P) -> Result<()> {
+        process_points(self, processor)
+    }
+}
+
+impl GeozeroGeometry for ListArray<i32> {
+    fn process_geom<P: GeomProcessor>(&self, processor: &mut P) -> Result<()> {
+        delegate_by_extension_type(self, processor)
+    }
+}
+
+impl GeozeroGeometry for ListArray<i64> {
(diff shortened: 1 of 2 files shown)


pka:
Code looks fine for me so far!

geozero has very limited support for empty (null) geometries. See this discussion for more details: https://github.com/georust/geozero/pull/20#discussion_r840003700

kylebarron:
For now I switched to prototyping a [geoarrow implementation as a crate in geopolars](https://github.com/geopolars/geopolars/tree/master/geopolars/geopolars-arrow) so that I could iterate faster, and in the future I think it would make sense to come back and figure out how to integrate with geozero.

I suppose I could also just implement `GeomProcessor` in the geoarrow crate and keep it as a separate implementation, like the flatgeobuf crate is.

kylebarron:
Closing this because this was implemented in https://github.com/geoarrow/geoarrow-rs

PR #80, opened 2022-11-23
WIP: GeoParquet reader

kylebarron:
I started to take a stab at this to resolve https://github.com/georust/geozero/issues/37 but ran into a few issues/questions:

- I put `process_geoarrow_feature_chunk` in the geoarrow mod, because the parquet reader's result is automatically geoarrow (of the wkb variant). In theory we could implement `GeozeroDatasource` for geoarrow directly as well to convert from an arrow table, but that might be less useful.
- It seemed not possible to write an impl for `GeozeroGeometry` in addition to `GeozeroDatasource`, because `GeozeroGeometry`'s `process_geom` only provides a non-mutable reader, and the parquet `FileReader` needed a mutable reference. Or maybe I'm missing something there.
- I'm not sure the best way of doing arrow -> geozero feature objects. Arrow2's "chunk" object is a vec of `Array` trait objects, so we need to do downcasting to get values out. Is it possible to do that downcasting outside of the loop so that we don't have to downcast on every iteration?
- I assume geozero has a record-only API? I.e. I can't process all the data column by column, it needs to be row-by-row?


pka:
Rebasing should solve the failed CI test.

nyurik:
I did the merge, but i have no domain knowledge about this PR

nyurik:
@kylebarron & @pka please update on this PR's path, if it needs more changes, or should be closed or merged

kylebarron:
Closing this because it's been (mostly) implemented in https://github.com/geoarrow/geoarrow-rs, with transitive bindings into geozero.

PR #87, opened 2022-12-12
geojson: escape quotes in written property name/values

brendanashworth:
As far as I can tell, serde-json automatically unescapes input, so this change is not needed on the reader.

Before this would output invalid JSON with random unescaped quotes in the property fields. Test fails before, now passes.


geozero/src/geojson/geojson_writer.rs
```diff
@@ -192,7 +192,10 @@ fn write_num_prop<'a, W: Write>(out: &'a mut W, colname: &str, v: &dyn Display)
 }
 
 fn write_str_prop<'a, W: Write>(out: &'a mut W, colname: &str, v: &dyn Display) -> Result<()> {
-    out.write_all(&format!(r#""{}": "{}""#, colname, v).as_bytes())?;
+    let esc_colname: String = colname.replace("\"", "\\\"");
+    let esc_value: String = format!("{}", v).replace("\"", "\\\"");
```
pka, line 196: I'm not sure, whether this additional String allocation is optimzied away by the compiler.
To be sure, I would escape directly in the `write_all` call.

pka (changes requested):
Thanks for pointing this out, makes absolutely sense. I've added a minor change request in the review comment.

geozero/src/geojson/geojson_writer.rs
brendanashworth: I wasn't sure how to escape `v` without using `format!()` because `v` is `Display` trait, not `str`. How would you do this without allocating a new string?

pka: So
`out.write_all(&format!(r#""{}": "{}""#, colname.replace("\"", "\\\""), v.replace("\"", "\\\"")).as_bytes())?;`
doesn't compile?

brendanashworth: Right, I get this:

```rust
error[E0599]: no method named `replace` found for reference `&dyn std::fmt::Display` in the current scope
   --> geozero/src/geojson/geojson_writer.rs:201:11
    |
201 |         v.replace("\"", "\\\"")
    |           ^^^^^^^ method not found in `&dyn std::fmt::Display`
```

Happy to check out other options if you know of any.

pka: This compiles:

```diff
 fn write_num_prop<'a, W: Write>(out: &'a mut W, colname: &str, v: &dyn Display) -> Result<()> {
-    out.write_all(&format!(r#""{}": {}"#, colname, v).as_bytes())?;
+    out.write_all(&format!(r#""{}": {}"#, colname.replace("\"", "\\\""), v).as_bytes())?;
     Ok(())
 }
 
-fn write_str_prop<'a, W: Write>(out: &'a mut W, colname: &str, v: &dyn Display) -> Result<()> {
-    out.write_all(&format!(r#""{}": "{}""#, colname, v).as_bytes())?;
+fn write_str_prop<'a, W: Write>(out: &'a mut W, colname: &str, v: &str) -> Result<()> {
+    out.write_all(
+        &format!(
+            r#""{}": "{}""#,
+            colname.replace("\"", "\\\""),
+            v.replace("\"", "\\\"")
+        )
+        .as_bytes(),
+    )?;
     Ok(())
 }
```

brendanashworth: Just pushed your suggested changes, thanks for figuring that out

brendanashworth:
Just changed the code as you suggested, thanks for reviewing @pka

brendanashworth:
Perhaps annoyingly, I didn't realize that there's more characters that need to be escaped in valid JSON, at least according to `jq`:

```
parse error: Invalid string: control characters from U+0000 through U+001F must be escaped at line ..., column ...
```

> In JSON the only characters you must escape are \, ", and control codes

[This blog post](https://lise-henry.github.io/articles/optimising_strings.html) seems to have some interesting suggestions on escaping things quickly so I'm going to try out an implementation, will follow up with another PR probably

PR #89, opened 2022-12-20
Adds diesel support for Ewkb type for postgis

myaple:
This additional allows for the seamless use of the Ewkb format in diesel to map to postgis geometry and geography types. Because postgis emits Ewkb for geometry types, we can make the wkb module an entrypoint into the GeoRust ecosystem and allow easy conversion to other types. While it is feasible to make any type that implements ToWkb and has a To* trait from wkb a base type in diesel, Ewkb is the most flexible because it is the base emission from postgis.


pka:
Thanks, code looks good! Could you check the failing test?

PR #93, opened 2022-12-29
derive FromSqlRow for Ewkb

sunny-g:
Am new to both `geozero` and `diesel`, but for the queries I was trying to write, I needed to add this derive to execute the following (truncated) code:

The diesel errors were kind of inscrutable, but from what I could surmise, the general issue is that while `st_union_agg` (which is a SQL function I defined to wrap the aggregate version of PostGIS' `ST_Union`) returns a `geozero::postgis::diesel::sql_types::Geometry`, an `Ewkb` cannot be obtained from the select statement without explicit conversion, selecting the `Geometry` column itself, or without this trait implementation.

Sorry I can't provide a better explanation.

```rust
#[derive(Queryable)]
struct CollectionDiff {
    pub start: DateTime<Utc>,
    pub end: DateTime<Utc>,
    pub updated: DateTime<Utc>,
    pub footprint: Ewkb,
}

stac::table
    .select((
        min(stac::observed).assume_not_null(),
        max(stac::observed).assume_not_null(),
        max(stac::updated).assume_not_null(),
        st_union_agg(stac::geometry),
    ))
    .get_result::<CollectionDiff>(&mut conn)?;

```

### DIFF
--- a/geozero/src/wkb/wkb_reader.rs
+++ b/geozero/src/wkb/wkb_reader.rs
@@ -7,7 +7,7 @@ use std::io::Read;
 #[cfg(feature = "with-postgis-diesel")]
 use crate::postgis::diesel::sql_types::{Geography, Geometry};
 #[cfg(feature = "with-postgis-diesel")]
-use diesel::expression::AsExpression;
+use diesel::{deserialize::FromSqlRow, expression::AsExpression};
 
 /// WKB reader.
 pub struct Wkb(pub Vec<u8>);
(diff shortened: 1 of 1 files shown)


michaelkirk:
I don't use diesel, but I'm a little surprised that you'd build an Ewkb from an entire row. I'd expect just a single field of the row to be the ewkb.

pka:
@myaple ?

myaple:
Sorry, complete oversight on my implementation. FromSqlRow allows you to use the FromSql implementation to convert a single value in a row (tuple) selection into the type. So, because all of my tests for this just used a struct instead of explicitly calling select on a tuple like @sunny-g did, I forgot to add the derive macro. My apologies @pka and thank you @sunny-g for putting this in.

pka:
Thanks @sunny-g (and @myaple for checking)!

sunny-g:
Thank you both! @pka @myaple

sunny-g:
@michaelkirk I'm not, this example is getting an `Ewkb` from the union of the individual column.

PR #103, opened 2023-02-26
Resolve some default-warn clippy lints in the geozero-shp crate

rye:
Some of the code of the `geozero-shp` crate was copied over from `shapefile-rs` in 08ef12816e3f3aa2b4a5b148b22927f1399cd0c6. As-is, that code has some clippy warnings in the `clippy::all` set. This makes hacking on the geozero-shp crate a bit cumbersome, as it's hard to just call `cargo clippy` and know whether your changes introduced new clippy lints.

The changes in this PR seem to — for the most part, at least — apply to those parts that were copied over. I didn't individually git-blame to verify this, but I didn't encounter anything geozero-specific in the changes I made.

I also intentionally kept this PR _very_ sparse (only making minimal changes so as to resolve clippy lints). I aim to follow up with another PR that does more cleanup, hopefully to include some more intense refactoring and full documentation.

### DIFF
--- a/geozero-shp/src/header.rs
+++ b/geozero-shp/src/header.rs
@@ -130,41 +130,38 @@ impl ShapeType {
 
     /// Returns whether the ShapeType has the third dimension Z
     pub fn has_z(self) -> bool {
-        match self {
-            ShapeType::PointZ
-            | ShapeType::PolylineZ
-            | ShapeType::PolygonZ
-            | ShapeType::MultipointZ => true,
-            _ => false,
-        }
+        matches!(
+            self,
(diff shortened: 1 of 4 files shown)


rye:
A couple of addenda:

- I am happy to add `#![deny(clippy::all)]` to `geozero-shp/src/lib.rs` to make the "default Rust style" strictly-enforced through errors.
- I didn't touch the other crates in the workspace. I am happy to do this as well!
  - I would even add `cargo clippy` to the CI workflow if that would be helpful.

pka:
Thanks a lot!
I'm fine with any clippy fix and a CI workflow would be helpful, indeed!

PR #109, opened 2023-03-12
More Clippy Fixes, Run Clippy in CI

rye:
This PR does a number of small changes to resolve clippy lints across the workspace. I'd recommend reviewing it commit-by-commit, since the overall diff is rather large.

A couple of notable things:

- This PR also enables Clippy to run in CI, to hopefully catch future regressions.
- I set up a clippy.toml for `geozero` to configure the "too-many-arguments" lint threshold to 8 (the default is 7) — this was done to avoid making breaking refactoring changes to the `GeomProcessor::coordinate` method. (I could just `#[allow(..)]` it instead if that makes more sense to reviewers.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -47,6 +47,12 @@ jobs:
           command: check
           args: --workspace --all-features --bins --examples --tests --benches
 
+      - name: run clippy
+        uses: actions-rs/cargo@v1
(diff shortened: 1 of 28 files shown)


pka:
This is great work, thanks! Maybe we should add a sentence about how to run clippy in Readme for ignorants like me?

PR #115, opened 2023-04-01
Update dependencies, 2021 edition

nyurik:
Bumped some versions to the more up-to-date ones, and fixed arrow2 array offset usage.
Updating `clap 3->4` will be a separate PR.  Updating `geo` to 0.24 has some compilation issues - not sure how to fix those yet.

### DIFF
--- a/geozero-bench/Cargo.toml
+++ b/geozero-bench/Cargo.toml
@@ -2,7 +2,7 @@
 name = "geozero-bench"
 version = "0.2.0"
 authors = ["Pirmin Kalberer <pka@sourcepole.ch>"]
-edition = "2018"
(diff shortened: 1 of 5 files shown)


michaelkirk:
Please do not merge this. 

Please read https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277

nyurik:
@michaelkirk thx, i keep forgetting about that article - the issue is that the versions in this repo are a mix of `major.minor` and `major.minor.patch` - so I would think it might make it better to keep it consistent?

Regardless, I reverted most of them to the original, and bumped just the ones i wanted to bump/fix

PR #131, opened 2023-04-05
Implement MVT tag builder

nyurik:
* Implement a new TagBuilder struct that accumulates keys and values with stable indexes, and once finished, can be used to serialize the MVT lookup tables.
* Implement TileValue enum to simplify and speed up access to the MVT values
* Some minor testing and other MVT cleanups
* Sort cargo features

### DIFF
--- a/geozero/src/mvt/tile_value.rs
+++ b/geozero/src/mvt/tile_value.rs
@@ -1,0 +1,92 @@
+use crate::mvt::tile::Value;
+use std::hash::Hash;
+
+/// A wrapper for the MVT value types.
+#[derive(Debug, Clone, PartialEq)]
+pub enum TileValue {
+    Str(String),
+    Float(f32),
+    Double(f64),
+    Int(i64),
+    Uint(u64),
+    Sint(i64),
+    Bool(bool),
+}
+
+impl From<TileValue> for Value {
+    fn from(tv: TileValue) -> Self {
+        match tv {
+            TileValue::Str(s) => Self {
+                string_value: Some(s),
+                ..Default::default()
+            },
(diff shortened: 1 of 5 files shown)


pka (approved):
:+1:

geozero/src/mvt/tile_value.rs
```diff
@@ -0,0 +1,92 @@
+use crate::mvt::tile::Value;
+use std::hash::Hash;
+
+/// A wrapper for the MVT value types.
+#[derive(Debug, Clone, PartialEq)]
+pub enum TileValue {
+    Str(String),
```
michaelkirk, line 7: I know this is already merged, but it wasn't open for long. I like this tag builder change - it cleans things up a bit. 

One nitpick: I feel like `TileValue::String` would be a better name, so as not to confuse it with rust's `str` type. Is there some reason to call it `Str`, or is just that it's shorter to type?

nyurik: Thx for feedback @michaelkirk! I actually initially called it `String` too, but the issue is that it is identical to the system string, making it impossible to use it without the enum qualifier like here:

https://github.com/georust/geozero/blob/1a1585a9fb9cfbcb3cc8d37280ab8b68746b739c/geozero/src/mvt/tag_builder.rs#L63-L66

So I opted to call it by a different name. I think I have seen this pattern elsewhere in the past, where a string wrapper is called `Str` with a different case to differentiate it while still indicate its close associativity with the string type.

michaelkirk: Fair point! I guess it's a trade off / matter of taste.

PR #136, opened 2023-04-08
PostgreSQL CI

nyurik:
* Run postgresql as part of the CI
* Refactor postgis tests a bit to consolidate connection code

### DIFF
--- a/geozero/tests/postgis.rs
+++ b/geozero/tests/postgis.rs
@@ -84,7 +84,10 @@ mod postgis_postgres {
         )?;
 
         let value: wkb::Decode<geos::Geometry> = row.get(0);
-        assert_eq!(value.geometry.unwrap().to_wkt().unwrap(), "POLYGON ((0.0000000000000000 0.0000000000000000, 2.0000000000000000 0.0000000000000000, 2.0000000000000000 2.0000000000000000, 0.0000000000000000 2.0000000000000000, 0.0000000000000000 0.0000000000000000))");
+        assert_eq!(
+            value.geometry.unwrap().to_wkt().unwrap(),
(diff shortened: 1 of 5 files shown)


nyurik:
@pka could you take a look at the [build log](https://github.com/georust/geozero/pull/136/checks) of this PR - I am not certain how PG tests should be executed.

pka:
The CI failures are at `std::env::var("DATABASE_URL").unwrap()`. In the CI Yaml it's `_URI` instead of `_URL`!

geozero/tests/postgis.rs
```diff
@@ -1,16 +1,34 @@
+mod pg {
+    fn get_db_string() -> String {
+        std::env::var("DATABASE_URL").unwrap()
+    }
+
+    #[cfg(feature = "with-postgis-postgres")]
+    pub fn get_conn() -> postgres::Client {
+        postgres::Client::connect(&get_db_string(), postgres::NoTls).unwrap()
```
pka, line 8: I would prefer using
`postgres::Client::connect(&get_db_string(), postgres::NoTls).unwrap()`
instead of `get_conn()` in the examples.
The goal of the tests are also being usage examples which should be usable with minimal adaption.

The `get_pool()` function is ok, since it makes the tests quite a lot shorter.

nyurik:
@pka thx, fixed the issue, but now 4 unit tests don't pass. I fixed two of them (it seems now the shown values don't have the unneeded `.000000000` ?  Could you take a look again pls

geozero/tests/postgis.rs
nyurik: fixed

pka (approved):

PR #147, opened 2023-06-02
RFC: Drop lifetime parameter of GeoJsonReader

adamreichold:
`&'a mut R` implements `Read` if `R` does so this should not reduce the set of supported types but it avoids forcing the user to provide a borrowed reference and allow taking ownership as well.


michaelkirk (approved):

pka:
> &'a mut R implements Read if R does

Good point, danke!

PR #149, opened 2023-06-22
Address a breaking change in a dependency, minor CI

nyurik:
* address a breaking change in geos lib - https://github.com/georust/geos/pull/133#issuecomment-1603243591
* make CI split fmt from clippy for simpler review


pka (approved):
:+1:

PR #152, opened 2023-08-08
Improve scaling code a bit

nyurik:
* On a second though, I think `to_mvt_unscaled` is more accurate. `raw` here would be confusing.
* There is no need to store bounds and do additional math ops on it. Can precompute a few things earlier on.
* Default can be auto-derived

P.S. @pka I think its better to push PRs to individual's forks, and only keep project-maintained versions in the primary repo. This way there would be no experimental branches in the main repo, reducing confusion.


nyurik:
P.S. Once you +1 this (or merge it into your branch), i can +1/merge yours into main

pka (approved):

PR #153, opened 2023-08-09
Add support for Spatialite and MySQL WKB dialects

Oreilles:
This PR adds support for [Spatialite](https://www.gaia-gis.it/gaia-sins/BLOB-Geometry.html) and [MySQL](https://dev.mysql.com/doc/refman/8.0/en/gis-data-formats.html) WKB dialects.

I'd happily discuss the implementation. Quirks of the spatialite dialect that required some refactoring:
- A geometry blob ends with `0xFE`. In order to know wether the `polygon_end` (or alike) is the last one, we keep track of the collection entity nesting level. If the nesting level is 0 when a geometry ends, it means it is the last one of the blob and the terminating byte `0xFE` must be appended. Even though `MULTI*` geometries cannot be part of a `GEOMETRYCOLLECTION` in Spatialite at the moment, this was my easiest go at the problem - and it also makes it more future-proof in case that feature is added at some point.
- In contrary to EWKB, the byte order and SRID info are not repeated in the nested entities headers, which means they have to be inherited from the parent entity info. In order to do that, I had to change the signature of the `read_header` parameter of `process_wkb_geom_n` so that the parent geometry info can be passed down to its children.
- The "compressed" polygon and linestring geometries stores their X,Y,Z coordinates as `f32` values, relative to the previous one (with the exception of the first and the last one), which required some refactoring to `process_coord` to avoid code duplication between the two implementations.

Side notes:
- Shouldn't `WkbWriter` implement `GeomProcessor::srid()` ?
- Only read support has been implemented for Compressed and TinyPoint geometries, write isn't implemented.
- Spatialite header stores the envelope as `[minx, miny, maxx, maxy]` (like GeoJSON) whereas gpkg stores it as `[minx, maxx, miny, maxy]`

I also refactored the roundtrip tests so they can be shared by the 5 dialects.


pka (changes requested):
Great addition, thanks a lot @Oreilles !

Please do a `cargo fmt` and we're good to merge.

Oreilles:
Thanks for the review, the latest commit fixes the formatting issues!

pka:
Thanks! Now we just have to make clippy happy as well. To test locally, you just have to run `cargo clippy` or to be sure with all options as in the CI: `cargo clippy --workspace --all-features --bins --examples --tests --benches`.

Oreilles:
I think it's good now!

geozero/src/wkb/wkb_reader.rs
```diff
@@ -293,17 +424,59 @@ pub(crate) fn process_wkb_geom_n<R: Read, P: GeomProcessor>(
     }
 }
 
+fn emit_coord<P: GeomProcessor>(
```
nyurik, line 427: I don't think this is a good approach here - you first create coordinate tuple on line 462, and then you discover you don't even need it if multi_dim is false.  I would inline this function

geozero/src/wkb/wkb_reader.rs
```diff
@@ -576,32 +771,168 @@ mod test {
...
         assert_eq!(
-            &ewkb_to_wkt("0111000000010000000400000000000000000000000000000000000000000000000000000000000000000022400000000000002240000000000000000000000000000000000000000000000000", false),
+            &wkb_to_wkt(WkbDialect::Ewkb, "0111000000010000000400000000000000000000000000000000000000000000000000000000000000000022400000000000002240000000000000000000000000000000000000000000000000", false),
             "TRIANGLE((0 0,0 9,9 0,0 0))"
         );
     }
 
-    fn ewkb_to_wkt(ewkb_str: &str, with_z: bool) -> String {
-        let ewkb = hex::decode(ewkb_str).unwrap();
+    #[test]
```
nyurik, line 790: I would leave the `ewkb_to_wkt` helper here - there is no point in changing all the unit tests, and makes it harder to spot any breaking changes.

geozero/src/wkb/wkb_common.rs
```diff
@@ -198,3 +200,12 @@ pub(crate) enum WKBByteOrder {
     Xdr = 0, // Big Endian
     Ndr = 1, // Little Endian
 }
+
+impl From<scroll::Endian> for WKBByteOrder {
```
nyurik, line 204: good addition, thanks!  I wonder if we need a reverse as well?

pka (approved):

geozero/src/wkb/wkb_reader.rs
```diff
@@ -70,29 +106,28 @@ pub fn process_wkb_type_geom<R: Read, P: GeomProcessor>(
...
 
 /// OGC WKB header.
 pub(crate) fn read_wkb_header<R: Read>(raw: &mut R) -> Result<WkbInfo> {
     let byte_order = raw.ioread::<u8>()?;
-    let endian = if byte_order == WKBByteOrder::Xdr as u8 {
-        scroll::BE
-    } else {
-        scroll::LE
-    };
+    let endian = Endian::from(byte_order != 0);
```
nyurik, line 130: sigh, `Endian::from(bool)` is not the most readable interface. I had to look up what `true` means :(

geozero/src/wkb/wkb_reader.rs
```diff
@@ -178,17 +215,111 @@ fn read_gpkg_header<R: Read>(raw: &mut R) -> Result<WkbInfo> {
...
+    let endian = Endian::from(flags & 0b0000_0001 != 0);
+    let is_tinypoint = flags & 0b1000_0000 != 0;
+
+    let srid = match raw.ioread_with::<i32>(endian)? {
+        0 => None,
+        val => Some(val),
+    };
+
+    let info = if is_tinypoint {
+        let envelope = Vec::<f64>::new();
```
nyurik, line 240: many of these lines should probably be inlined for ease-of-reading. As long as the `raw.ioread_with...` is in the same order, it seems like it should be ok.  Also, i think `matches!(...)` is a bit hard to read -- perhaps a simple bool is better, e.g. `(type_id_dim == 2 || type_id_dim == 4)` ?  Or is this something clippy is having too much fun with?

nyurik:
lol, seems like @pka and I were reviewing it at the same time...

nyurik (dismissed):
see inline

geozero/src/wkb/wkb_reader.rs
nyurik: P.S. this one is not exactly actionable - just a note in case you can think of a cleaner API usage here

geozero/src/wkb/wkb_reader.rs
Oreilles: I personally find the `matches!(...)` macro easier on the eyes, but I'm fine with either - I don't think clippy has any problem with one solution or the other.

geozero/src/wkb/wkb_reader.rs
Oreilles: Yup, I agree that `Endian::from(bool)` is not the very explicit at first glance... If we implement the reverse of `From<scroll::Endian> for WKBByteOrder` as you suggested, then maybe we could also implement `From<u8> for WKBByteOrder` and chain the two. That's a lot of code for basically a bool check though so...

nyurik: yeah, probably not worth the complexity - maybe just add an extra line to make it more readable, like this?

```rust
let is_little_endian = byte_order != 0;
let endian = Endian::from(is_little_endian);
```

Oreilles: Much better, I'll do that 👍

geozero/src/wkb/wkb_reader.rs
Oreilles: Fixed with latest commit

geozero/src/wkb/wkb_reader.rs
```diff
@@ -70,29 +106,29 @@ pub fn process_wkb_type_geom<R: Read, P: GeomProcessor>(
...
 
 /// OGC WKB header.
 pub(crate) fn read_wkb_header<R: Read>(raw: &mut R) -> Result<WkbInfo> {
     let byte_order = raw.ioread::<u8>()?;
-    let endian = if byte_order == WKBByteOrder::Xdr as u8 {
-        scroll::BE
-    } else {
-        scroll::LE
-    };
+    let is_little_endian = byte_order != 0;
```
nyurik, line 130: turns out there is a much cleaner solution, but it requires a bit of research.  https://docs.rs/scroll/latest/scroll/ has a lot of docs for that, but essentially instead of using `raw.ioread::<u8>()` you can use `raw.ioread::<WKBByteOrder>`. In order for that to work, you would have to implement `impl IntoCtx<scroll::Endian> for WKBByteOrder {` or something similar (I never used this crate, just saw some docs about cread vs gread vs pread - not sure which is which).  At the end of the day, it should offer a very easy way to just do all the on-the-fly conversions cleanly

nyurik: P.P.S. note that we don't need to fix ALL such issues at the same time, can be done in another PR.  For more docs on this - https://github.com/m4b/scroll#stdio-api

geozero/src/wkb/wkb_reader.rs
Oreilles: Reverted this change

geozero/src/wkb/wkb_writer.rs
```diff
@@ -154,6 +157,68 @@ impl<'a, W: Write> WkbWriter<'a, W> {
...
+            WkbDialect::Ewkb => Ok(()),
+            WkbDialect::Geopackage => Ok(()),
+            WkbDialect::MySQL => Ok(()),
+            WkbDialect::SpatiaLite => {
+                if self.nesting_level == 0 {
+                    self.out.iowrite::<u8>(0xFE)?;
+                }
+                Ok(())
+            }
+        }
```
nyurik, line 220:
```suggestion
    match self.dialect {
        WkbDialect::SpatiaLite => {
            if self.nesting_level == 0 {
                self.out.iowrite::<u8>(0xFE)?;
            }
        }
        WkbDialect::Wkb | WkbDialect::Ewkb | WkbDialect::Geopackage | WkbDialect::MySQL => {},
    }
    Ok(())
```

geozero/src/wkb/wkb_writer.rs
```diff
@@ -224,45 +304,76 @@ impl<W: Write> GeomProcessor for WkbWriter<'_, W> {
     }
     fn polygon_end(&mut self, _tagged: bool, _idx: usize) -> Result<()> {
         self.geom_state = GeomState::Normal;
+        self.write_footer()?;
         Ok(())
```
nyurik, line 308:
```suggestion
        self.write_footer()
```

geozero/src/wkb/wkb_writer.rs
```diff
@@ -224,45 +304,76 @@ impl<W: Write> GeomProcessor for WkbWriter<'_, W> {
...
+        self.nesting_level += 1;
         self.write_header(WKBGeometryType::GeometryCollection)?;
         self.out.iowrite_with(size as u32, self.endian)?;
         Ok(())
     }
-
+    fn geometrycollection_end(&mut self, _idx: usize) -> Result<()> {
+        self.nesting_level -= 1;
+        self.write_footer()?;
+        Ok(())
```
nyurik, line 329:
```suggestion
        self.write_footer()
```

geozero/src/wkb/wkb_writer.rs
```diff
@@ -224,45 +304,76 @@ impl<W: Write> GeomProcessor for WkbWriter<'_, W> {
...
     fn multicurve_begin(&mut self, size: usize, _idx: usize) -> Result<()> {
+        self.nesting_level += 1;
         self.write_header(WKBGeometryType::MultiCurve)?;
         self.out.iowrite_with(size as u32, self.endian)?;
         Ok(())
     }
+    fn multicurve_end(&mut self, _idx: usize) -> Result<()> {
+        self.nesting_level -= 1;
+        self.write_footer()?;
+        Ok(())
```
nyurik, line 364:
```suggestion
        self.write_footer()
```

geozero/src/wkb/wkb_writer.rs
```diff
@@ -224,45 +304,76 @@ impl<W: Write> GeomProcessor for WkbWriter<'_, W> {
...
+        self.nesting_level += 1;
         self.write_header(WKBGeometryType::MultiSurface)?;
         self.out.iowrite_with(size as u32, self.endian)?;
         Ok(())
     }
-
+    fn multisurface_end(&mut self, _idx: usize) -> Result<()> {
+        self.nesting_level -= 1;
+        self.write_footer()?;
+        Ok(())
```
nyurik, line 375:
```suggestion
        self.write_footer()
```

geozero/src/wkb/wkb_writer.rs
```diff
@@ -271,18 +382,25 @@ impl<W: Write> GeomProcessor for WkbWriter<'_, W> {
     }
     fn triangle_end(&mut self, _tagged: bool, _idx: usize) -> Result<()> {
         self.geom_state = GeomState::Normal;
+        self.write_footer()?;
         Ok(())
```
nyurik, line 386:
```suggestion
        self.write_footer()
```

geozero/src/wkb/wkb_writer.rs
```diff
@@ -292,114 +410,157 @@ impl<W: Write> FeatureProcessor for WkbWriter<'_, W> {}
...
+        assert!(process_wkb_type_geom(&mut wkb_in.as_slice(), &mut writer, dialect).is_ok());
+
+        assert_eq!(hex::encode(wkb_in), hex::encode(wkb_out));
     }
 
     #[test]
     fn ewkb_geometries() {
         // SELECT 'POINT(10 -20)'::geometry
-        assert!(ewkb_roundtrip(
+        roundtrip(
```
nyurik, line 437: all these unit test changes - probably not worth changing - just keep the helper method `ewkb_roundtrip`

nyurik (commented):
a few minor nits, but looks almost ready to merge. Thanks for all the patience!! :)

Oreilles:
> Thanks for all the patience!! :)

Thanks for your time, it's very pleasing contributing to open source projects with responsive maintainers 👍

geozero/src/wkb/wkb_writer.rs
Oreilles: Then we'd have four functions `ewkb_roundtrip`, `gpkg_roundtrip`, `spatialite_roundrip` and `mysql_roundtrip` with identical code..?

nyurik: ah, i see, ok, nvm, lets modify the tests

nyurik:
please fix the conflict and lets merge

Oreilles:
Solved

nyurik:
At first i made a few minor comments here, but then just created a PR for your PR - please merge if agree -- https://github.com/Oreilles/geozero/pull/1

Oreilles:
I merge the two PR!
On an unrelated side, I have a question about the processor functions `xy` and `coordinate`: since all parameter of `coordinate` but `x` and `y` are of type `Option`, would it be possible to only use `coordinate` and let the writer decide what to do with them ?
My concern is that if you parsed for example a `Wkb` blob with XY coordinates, and process it with a `WkbWriter` for which you set `dims = CoordDimensions::xyzm()`, then the generated WKB would be invalid: its header would indicate the presence of `XYZM` values, but since only `processor.xy` would be called by the reader (since the input doesn't have ZM), the ZM coordinates bytes wouldn't be written as they should.
Having the readers always call `processor.coordinate` and let the processors decide what to do with the unnecessary (or in this example, absent) coordinate values would mitigate this issue.
I picked a `Wkb` to `Wkb` example but that behavior seems to be shared by other readers as well.

pka:
> On an unrelated side, I have a question about the processor functions `xy` and `coordinate`: since all parameter of `coordinate` but `x` and `y` are of type `Option`, would it be possible to only ever use `coordinate` and let the writer decide what to do with them ?

The reason for having two methods was performance. This method is the innermost loop of any geometry operation (except for points) and passing additional optional params costs performance for the most common xy case.

> My concern is that if you parsed for example a Wkb blob with XY coordinates, and process it with a WkbWriter for which you set dims = CoordDimensions::xyzm(), then the generated WKB would be invalid

Your concerns are valid, if the reader implementing the `GezeroGeometry` trait doesn't look at the requested dimensions.
Code example for dimension handling: https://github.com/georust/geozero/blob/ece8dfe4955c2b161e8dd001431bd38cc85a8e76/geozero/src/geojson/geojson_reader.rs#L238-L251
`WkbReader` looks also fine, but there are maybe readers which don't respect `GeomProcessor::dimensions` resp. `GeomProcessor::multi_dim`, which would be a bug.

In your WKB xy to WKB xyzm example, I don't think that the result is invalid, you just loose the additional dimensions.

Oreilles:
> In your WKB xy to WKB xyzm example, I don't think that the result is invalid, you just loose the additional dimensions.

If we process a WKB that doesn't have ZM coordinates, it will only call `processor.xy`:

https://github.com/georust/geozero/blob/ece8dfe4955c2b161e8dd001431bd38cc85a8e76/geozero/src/wkb/wkb_reader.rs#L315-L319

And then only X and Y will be written:

https://github.com/georust/geozero/blob/ece8dfe4955c2b161e8dd001431bd38cc85a8e76/geozero/src/wkb/wkb_reader.rs#L315-L319

When a parser will try to read the WKB emitted, il will try to read four `f64` values for each point, but only two would have been written.

> The reason for having two methods was performance. This method is the innermost loop of any geometry operation (except for points) and passing additional optional params costs performance for the most common xy case.

You might have tested it already (I haven't) but in this case, maybe the branching caused by having to handle four different outcomes outweight the performance gain of only calling the method with the least amout of parameters:
- The reader emitted XY, and the writer have to write XY
- The reader emitted XY, but the writer have to write XYZM
- The reader emitted XYZM, but the writer have to write XY
- The reader emitted XYZM, and the writer have to write XYZM

nyurik:
@Oreilles i think this one is ready to go once conflict is fixed.  One thing I forgot to ask about the other PR (which affect this PR as well) -- why do you need to call `processor.srid(info.srid)?;` everywhere?  Shouldn't the processor be created with a given srid once?

Oreilles:
@nyurik calling processor.srid(info.srid) allowed for inferring the EWKT geometry srid from the WKB blob without having to manyally specify it. The trait `ToWkt` has a `to_ewkt` function that have an optional srid argument for when the source Wkb doesn't have one, but you still want one in the output. If the input Wkb has an srid defined, then this parameter can be omited and the one provided by the reader will be used (which is why I had to call `processor.srid(info.srid)`. I wrote this so that the srid provided in the `to_ewkt` function parameter takes predominance over the one provided by the reader: 

https://github.com/georust/geozero/blob/5e92027df76e2d0a3e29495dcc51047400cb226c/geozero/src/wkt/wkt_writer.rs#L82-L85

nyurik:
Awesome work, thanks!!

PR #155, opened 2023-08-10
Adds Ewkt dialect

Oreilles:
This PR adds:
- `enum WktDialect` with options `WktDialect::Wkt` and `WktDialect::Ewkt` to `wkt` module
- `struct EwktString(pub String)`
- `fn to_ewkt(&self, srid: Option<i32>) -> Result<String>` to trait `ToWkt`
- `fn to_wkt_dialect(&self, dialect: WktDialect, dims: CoordDimensions, srid: Option<i32>) -> Result<String>` to trait `ToWkt`

The trait `FromWkb` is implemented for `EwktString` so that the srid is inferred from the Wkb stream if available. The later is made possible by implementing `GeomProcessor::srid` for `WktWriter` and calling `processor.srid(info.srid)` in `process_wkb_geom`.

Side notes:
- This PR only implements Ewkt write. Ewkt read should probably be implemented in the `wkt` crate.
- Should we add `ProcessToCsv::to_csv_ewkt` and/or `ProcessToCsv::to_csv_dialect`?


nyurik:
TBH, I am not too happy about changing all of the APIs to force the user to always be WKT dialect aware. I would much rather keep the existing API, plus perhaps add a new constructor, e.g. keep the `WktWriter::new(w: impl Write)`, and add another one like `WktWriter::with_dialect(w: impl Write, dialect: WktDialect)`?

Oreilles:
Sure, I changed the signature of the constructor for the sake of consistency with `WkbWriter` but I would definetly understand that an API change would not be acceptable.

nyurik:
oh, i didn't realize we already did it that way with WkbWriter... I'm a bit on a fence on this one then - ideally I would want Wkb to always done the same way - `WkbWriter::with_dialect(...)`, but I will let @pka to chime in on this one

pka (changes requested):
I'm fine with changing the `WktWriter`-API. The `ToWkt` trait API should be kept compatible if possible, which is the case here.

Comments:
* We should use the opportunity to drop the lifetime in `ẀktWriter`, like in #147
* Run `cargo fmt`

Oreilles:
Thanks for the review, I ran `fmt` and took example on the linked PR to remove the `WktReader` lifetime 👍

nyurik:
@Oreilles thx!  Do you need help merging main branch into your PRs?

Oreilles:
I just rebased onto `main` and solved the few conflicts.

geozero/src/wkt/mod.rs
```diff
@@ -9,25 +9,46 @@ pub use wkt_writer::*;
...
     /// Convert to WKT.
     pub trait ToWkt {
         /// Convert to 2D WKT String.
         fn to_wkt(&self) -> Result<String>;
+        /// Convert to EWKT String.
+        fn to_ewkt(&self, srid: Option<i32>) -> Result<String>;
         /// Convert to WKT String with dimensions.
         fn to_wkt_ndim(&self, dims: CoordDimensions) -> Result<String>;
+        /// Convert to WKT String with srid, dimensions and dialect.
+        fn to_wkt_dialect(
```
nyurik, line 24: Technically this is more of a `to_wkt_with_options` or `to_wkt_extended` method, not just `dialect`... Naming is hard, we may want to bikeshed this one.... (note that I am still kinda uneasy about the `new` constructor now requiring an additional dialect param even though most of the users probably won't ever care about it, and plus it doesn't allow us to have a default

geozero/src/wkt/mod.rs
```diff
@@ -40,18 +61,45 @@ pub(crate) mod conversion {
...
+#[derive(PartialEq, Debug, Clone, Copy)]
+pub enum WktDialect {
+    Wkt,
+    Ewkt,
+}
+
+impl Default for WktDialect {
+    fn default() -> Self {
+        WktDialect::Wkt
+    }
```
nyurik, line 104:
```suggestion
#[derive(Default, PartialEq, Debug, Clone, Copy)]
pub enum WktDialect {
    #[default]
    Wkt,
    Ewkt,
```

pka:
After discussing the PR with @nyurik, I'm also for keeping the `WktWriter::new` signature and adding a second method `WktWriter::with_dialect`. For `WkbWriter` the dialect is essential, but WKT users usually don't care and I don't want to force them to think about it. What's your use case of using EWKT`? Can't remember anyone mentioning it for a long time.

nyurik (commented):
I made a few suggestions in code.  I think we shouldn't be removing the lifetime in the same PR - lets just figure out the dialects first, as it is not certain?

Oreilles:
> After discussing the PR with @nyurik, I'm also for keeping the WktWriter::new signature and adding a second method WktWriter::with_dialect. For WkbWriter the dialect is essential, but WKT users usually don't care and I don't want to force them to think about it. 

That's understandable, I'll do the modification 👍

> What's your use case of using EWKT ?

Even though it's not a standard, thanks to PostGIS success it's still quite common and (AFAIAA) is the only human-friendly geometry format that allows SRID declaration. It will be used as a user input format in a product I'm working on.

geozero/src/wkt/mod.rs
Oreilles: This is the first Rust project I'm contributing to and don't know all the niceties yet, thanks for showing me!

geozero/src/gdal/gdal_reader.rs
```diff
@@ -183,16 +183,7 @@ mod test {
...
-            geo.to_wkt_ndim(CoordDimensions {
-                z: true,
-                m: false,
-                t: false,
-                tm: false
-            })
-            .unwrap(),
-            wkt
-        );
+        assert_eq!(geo.to_ewkt(None).unwrap(), wkt);
```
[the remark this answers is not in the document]
nyurik: thx, yeah, lets minimize test changes to make sure we don't break things too badly :)

geozero/src/wkt/mod.rs
nyurik: I just hope the number of dialects don't explode :)  `to_ewkt` and `to_ewkt_ndim` sounds fine.

geozero/src/wkt/mod.rs
```diff
@@ -40,18 +61,40 @@ pub(crate) mod conversion {
...
-    use crate::wkt::{WktString, WktWriter};
+    use crate::wkt::{EwktString, WktDialect, WktString, WktWriter};
+    use crate::CoordDimensions;
     use std::io::Read;
 
     impl FromWkb for WktString {
         fn from_wkb<R: Read>(rdr: &mut R, dialect: WkbDialect) -> Result<Self> {
             let mut out: Vec<u8> = Vec::new();
-            let mut writer = WktWriter::new(&mut out);
+            let mut writer = WktWriter::with_dialect(&mut out, WktDialect::Wkt);
```
nyurik, line 71: shouldn't this keep using `new()`?

geozero/src/wkt/mod.rs
```diff
@@ -40,18 +61,40 @@ pub(crate) mod conversion {
...
             Ok(WktString(wkt))
         }
     }
+
+    impl FromWkb for EwktString {
+        fn from_wkb<R: Read>(rdr: &mut R, dialect: WkbDialect) -> Result<Self> {
+            let mut out: Vec<u8> = Vec::new();
+            let mut writer = WktWriter::with_dialect(&mut out, WktDialect::Ewkt);
+            writer.dims = CoordDimensions::xyzm();
+            crate::wkb::process_wkb_type_geom(rdr, &mut writer, dialect)?;
```
nyurik, line 85: not sure if actionable, just checking if this is correct: both the writer and the processor need a dialect?   And when reading, the processor cannot auto-detect the dialect, or can it?

Oreilles: So an `Ewkt` string can be parsed from several `Wkb` dialects, which is why we need to pass it as a parameter. In the current implementation, no detection of the `Wkb` dialect is implemented; I guess it would be pretty tricky to do so, as they all have their own quirks.

geozero/src/wkt/mod.rs
nyurik: sigh :)  I think we may as well just use `to_wkt_with_opts`, and provide all those detalis. We may even want to get rid of the `to_wkt_ndims` to reduce the api surface (TBD)

I just really don't like the `to_wkt_dialect` because it actually has many more params than just the dialect one, making it confusing

nyurik (commented):
seems like this one is ready to go once we settle on the `to_wkt_X` name...  I think `to_wkt_with_opts` is the cleanest?

geozero/src/wkt/mod.rs
Oreilles: I agree, not sure what the best solution is here so I'll that one up to you..!

nyurik:
I made a few more suggestions for this PR in https://github.com/Oreilles/geozero/pull/2

* rename `to_wkt_dialect` -> `to_wkt_with_opts`
* make `WktWriter` content private!  It's ridiculous to have write access to a writer from outside
* introduce `WktWriter::with_dims` to allow creation of a writer with the needed params

pka (approved):

Issue #156, opened 2023-08-12
Make srid & dim fields of all writers private, and add ctor


nyurik:
As was done in a recently merged [here](https://github.com/Oreilles/geozero/pull/2/files#diff-553c2186acf21c8c360979841037ea9343c11e4e2937438cb9eca2c20d0d08cdR9) as part of https://github.com/georust/geozero/pull/155 for `WktWriter`, we need to remove `pub` on `srid` and `dim` fields, making it an internal implementation detail. Instead, there should be an extra writer constructor, e.g. `WktWriter::with_opts(&mut out, dialect, dims, srid)` that takes all the needed options.

This makes access cleaner, and removes a potential bug of dynamically changing srid or dimensions during processing. Unless of course the writer should support their dynamic modification, but even in that case it should be a function, not external variable access.

PR #158, opened 2023-08-12
Improve WkbWriter constructors

pka:
This PR adresses #156 (first commit) and should fix different dimensions in reader and writer (second commit).

Follow-up to discussion with @Oreilles in https://github.com/georust/geozero/issues/153#issuecomment-1675943714

Instead of the ugly `with_extended_opts` constructor we should probably use a builder pattern.

### DIFF
--- a/geozero/src/gpkg/geopackage.rs
+++ b/geozero/src/gpkg/geopackage.rs
@@ -53,9 +53,13 @@ impl<T: GeozeroGeometry + Sized> sqlx::Type<Sqlite> for wkb::Encode<T> {
 impl<'q, T: GeozeroGeometry + Sized> Encode<'q, Sqlite> for wkb::Encode<T> {
     fn encode_by_ref(&self, args: &mut Vec<SqliteArgumentValue<'q>>) -> IsNull {
         let mut wkb_out: Vec<u8> = Vec::new();
-        let mut writer = wkb::WkbWriter::new(&mut wkb_out, wkb::WkbDialect::Geopackage);
-        writer.dims = self.0.dims();
-        writer.srid = self.0.srid();
+        let mut writer = wkb::WkbWriter::with_opts(
+            &mut wkb_out,
+            wkb::WkbDialect::Geopackage,
+            self.0.dims(),
+            self.0.srid(),
+            Vec::new(),
+        );
         self.0
             .process_geom(&mut writer)
             .expect("Failed to encode Geometry");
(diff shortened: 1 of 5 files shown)


nyurik:
I think this is a good direction, thx! Would it make sense to perhaps simplify all readers/writers to own underlying `impl Read`/`impl Write`?  Same as `std::io::BufWriter` and others?  The wrappers would still allow access to the underlying `inner` reader/writer, same as buffer - with `get_ref` and `get_mut`

Oreilles:
Thanks for solving this, I was about to go at it when I saw your PR 🙂

Regarding the constructor issue: I don't know if that's a common pattern in rust, but in Typescript you often create functions accepting a single parameter object containing all the options, rather than one function argument for each parameter. It makes it easier to merge default values and reduce friction with adding new options. It has the disadvantage of being a bit more verbose though.

In this case, it could look something like this (untested):

```rust
#[derive(Default)]
pub struct WkbWriterParams {
    pub dialect: WkbDialect,
    pub dims: CoordDimensions,
    pub srid: Option<i32>,
    pub envelope: Vec<f64>,
    pub envelope_dims: CoordDimensions,
    pub extended_gpkg: bool,
}

impl<'a, W: Write> WkbWriter<'a, W> {
    pub fn new(out: &'a mut W, params: WkbWriterParams) -> WkbWriter<'a, W> {
        WkbWriter {
            dialect: params.dialect,
            dims: params.dims,
            srid: params.srid,
            envelope: params.envelope,
            envelope_dims: params.envelope_dims,
            extended_gpkg: params.extended_gpkg,
            empty: false,
            endian: scroll::LE,
            dialect,
            first_header: true,
            geom_state: GeomState::Normal,
            out,
        }
    }
}
```

which you would then call as such:
```rs
let writer = WkbWriter::new(out, WkbWriterParams { srid: Some(4326), Default::default() })
```

nyurik:
Rust usually uses builder pattern as it is far more concise and does not break when options are changed, e.g. [this guide](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html).  The biggest unknown is if the builder should return `&mut Self` or `Self` on each setter

pka:
I'm hesitating to put too much work into this mostly internal API. We could also implement `Default` for writer structs instead of having `with_extended_opts` or builder functions. But implementing`Default` isn't easy because of `out: &'a mut W`. So yes, we should probably switch to `out: W`.

michaelkirk:
> So yes, we should probably switch to out: W. 

This is done in #162

nyurik (approved):

PR #160, opened 2023-08-13
Fix invalid WKT empty geometries

Oreilles:
Currently, empty geometries would be writter in WKT as:
```
LINESTRING()
POLYGON()
GEOMETRYCOLLECTION()
```
which isn't valid WKT.

This PR addresses this issue and output instead:
```
LINESTRING EMPTY
POLYGON EMPTY
GEOMETRYCOLLECTION EMPTY
```

### DIFF
--- a/geozero/src/wkt/wkt_writer.rs
+++ b/geozero/src/wkt/wkt_writer.rs
@@ -76,6 +76,8 @@ impl<'a, W: Write> WktWriter<'a, W> {
             if geometry_size > 0 {
                 self.out.write_all(b")")?;
             }
+        } else {
+            debug_assert!(false, "ended geometry that didn't start");
         }
         Ok(())
     }
--- a/geozero/src/wkt/wkt_writer.rs
+++ b/geozero/src/wkt/wkt_writer.rs
@@ -6,32 +6,32 @@ use std::vec;
 use super::WktDialect;
 
 /// WKT Writer.
-pub struct WktWriter<'a, W: Write> {
+pub struct WktWriter<W: Write> {
     dims: CoordDimensions,
     srid: Option<i32>,
     dialect: WktDialect,
     first_header: bool,
     /// Stack of in-progress geometry sizes
     geometry_sizes: Vec<usize>,
-    out: &'a mut W,
+    pub(crate) out: W,
 }
 
-impl<'a, W: Write> WktWriter<'a, W> {
-    pub fn new(out: &'a mut W) -> WktWriter<'a, W> {
+impl<W: Write> WktWriter<W> {
+    pub fn new(out: W) -> Self {
         Self::with_opts(out, WktDialect::Wkt, CoordDimensions::default(), None)
     }
 
-    pub fn with_dims(out: &'a mut W, dims: CoordDimensions) -> WktWriter<'a, W> {
+    pub fn with_dims(out: W, dims: CoordDimensions) -> Self {
         Self::with_opts(out, WktDialect::Wkt, dims, None)
     }
 
     pub fn with_opts(
-        out: &'a mut W,
+        out: W,
         dialect: WktDialect,
         dims: CoordDimensions,
         srid: Option<i32>,
-    ) -> WktWriter<'a, W> {
-        WktWriter {
+    ) -> Self {
+        Self {
             dims,
             srid,
             dialect,
@@ -83,7 +83,7 @@ impl<'a, W: Write> WktWriter<'a, W> {
     }
 }
 
-impl<W: Write> GeomProcessor for WktWriter<'_, W> {
+impl<W: Write> GeomProcessor for WktWriter<W> {
     fn dimensions(&self) -> CoordDimensions {
         self.dims
     }
@@ -216,9 +216,9 @@ impl<W: Write> GeomProcessor for WktWriter<'_, W> {
     }
 }
 
-impl<W: Write> PropertyProcessor for WktWriter<'_, W> {}
+impl<W: Write> PropertyProcessor for WktWriter<W> {}
 
-impl<W: Write> FeatureProcessor for WktWriter<'_, W> {}
+impl<W: Write> FeatureProcessor for WktWriter<W> {}
 
 #[cfg(test)]
 mod test {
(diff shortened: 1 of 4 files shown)


geozero/src/wkt/wkt_writer.rs
```diff
@@ -53,23 +57,26 @@ impl<'a, W: Write> WktWriter<'a, W> {
...
+        if size == 0 {
+            self.out.write_all(b" EMPTY")?;
         } else {
             self.out.write_all(b"(")?;
         }
         Ok(())
     }
     fn geom_end(&mut self) -> Result<()> {
-        self.out.write_all(b")")?;
+        if let Some(geometry_size) = self.geometry_sizes.pop() {
```
michaelkirk, line 75: Would we ever expect this to be `None`?

michaelkirk (approved):
Good catch!

geozero/src/wkt/wkt_writer.rs
```diff
@@ -112,96 +119,95 @@ impl<W: Write> GeomProcessor for WktWriter<'_, W> {
     }
 
     fn empty_point(&mut self, idx: usize) -> Result<()> {
-        self.geom_begin(idx, b"POINT EMPTY")
-        // we intentionally omit calling geom_end(), because POINT EMPTY has no closing paren
+        self.geom_begin(b"POINT", true, 0, idx)?;
+        self.geom_end()
```
michaelkirk, line 123: This seems less weird. 👍

geozero/src/wkt/wkt_writer.rs
Oreilles: A geometry would have to end before it began, so in theory no ?

michaelkirk: Consider adding something like `else { debug_assert!(false, "ended geometry that didn't start") }` as a "this should never happen but we don't want to blow up in production if it does"

Oreilles: Done 👍

Oreilles:
We have a problem with the csv `BufferingWktWriter`... since it instanciate a new writer for every processor function call, the state isn't persisted and the implementation cannot work.

geozero/src/wkt/wkt_writer.rs
Oreilles: Nice catch... https://github.com/georust/geozero/pull/160#issuecomment-1676431784

michaelkirk:
> We have a problem with the csv BufferingWktWriter... since it instanciate a new writer for every processor function call, the state isn't persisted and the implementation cannot work.

I've just opened https://github.com/Oreilles/geozero/pull/3 against your own branch that I think addresses this (and simplifies my old crappy code from yesteryear!). Take a look and merge into your branch if you think it makes sense.

michaelkirk:
> If we had access to the geometry size in processor functions polygon_end etc., then we wouldn't have to manage the geometry_sizes stack 🤔

I'm also open to this approach (or we could do both) but it seems like a bigger breaking change, so my preference at this point would be to avoid it.

geozero/src/wkt/wkt_writer.rs
```diff
@@ -1,38 +1,42 @@
...
-    out: &'a mut W,
+    /// Stack of in-progress geometry sizes
+    geometry_sizes: Vec<usize>,
+    pub(crate) out: W,
 }
 
-impl<'a, W: Write> WktWriter<'a, W> {
-    pub fn new(out: &'a mut W) -> WktWriter<'a, W> {
+impl<W: Write> WktWriter<W> {
+    pub fn new(out: W) -> Self {
```
michaelkirk, line 20: This LGTM, but wanted to highlight the change for other reviewers that we're changing WktWriter to take any `impl Write` rather than only `&mut impl Write`.

nyurik: I'm all for simplifying writters so that they own the underlying `impl Write` - but should it possibly be a separate PR that changes them all?  It feels like a counter-pattern to introduce it in just one writer as part of an unrelated empty geometry support. Then we can rebase this PR on top of it?

pka: I'm fine with doing it for all writers. Who wants?

nyurik: I'm ok either way, just as long as we don't release until all writers are more-or-less lined up (and of course its a minor rather than a patch change

michaelkirk: > and of course its a minor rather than a patch change

Just to be clear, changing from `&mut Write -> Write` is a non-breaking change, right?

michaelkirk: Yes, there is a breaking change to WktWriter in this PR if that's what you mean (see the changelog entry).

So is everyone OK to merge this?

nyurik (approved):
lgtm

nyurik:
i thought there were some discussions in rust itself to stop using bors because github now covers most such usecases?

michaelkirk:
Yes, there were some of us trying to figure out GH merge queues a couple weeks ago. If you look at the proj crate’s recent commits for example, you’ll see some of my misadventures. As far as I know we don’t have any working examples on georust where GH’s merge queues function like bors: i.e. I trigger the process, then the tests run against an up-to-date branch and only get merged into main once they pass.It seems so basic, but i kept finding that the branches would merge immediately, regardless of if the tests pass.I’m sure it’s possible, I just didn’t personally dedicate the time to figuring it out. If you tell me what series of buttons to press I can press them. On Aug 16, 2023, at 12:57, Yuri Astrakhan ***@***.***> wrote:﻿
i thought there were some discussions in rust itself to stop using bors because github now covers most such usecases?

michaelkirk:
I just realized that this project does not, and never did, use bors. 😅

nyurik:
@michaelkirk in Martin, we have branch protection rule for `main`:
* Require a pull request before merging
* Require approvals (1)
* Require status checks to pass before merging (with most of the key CI jobs listed as required)

I think if you also enable `Require branches to be up to date before merging`, you will achieve what you want:  a way to +1 something, plus a way to say "squash once CI passes" - this way it will not be squashed unless everything is good, and it will not pass if another PR merges ahead of it.  Granted that it will fail if another PR merges first though.

michaelkirk:
I got it working! Ultimately I think the issue was [something else](https://github.com/georust/proj/pull/179#issuecomment-1684521823), but thanks for the sanity check @nyurik.

PR #161, opened 2023-08-13
Implement `empty_point` for `GeoJsonWriter`

Oreilles:
According to the RFC, empty points may be represented with empty coordinates:
```
RFC 7946


3.1.  Geometry Object

   o  A GeoJSON Geometry object of any type other than "GeometryCollection" has a member with the name "coordinates". The value of the "coordinates" member is an array.  The structure of the elements in this array is determined by the type of geometry.  GeoJSON processors MAY interpret Geometry objects with empty "coordinates" arrays as null objects.
```
This PR therefore implements the GeometryProcessor method `point_empty` for `GeojsonWriter` by writing a point with empty coordinates.
For reference, PostGIS also has this behavior:

```sql
SELECT ST_AsGeoJSON('POINT EMPTY'::geometry);
--> {"type":"Point","coordinates":[]}

### DIFF
--- a/geozero/src/geojson/geojson_writer.rs
+++ b/geozero/src/geojson/geojson_writer.rs
@@ -98,6 +98,12 @@ impl<W: Write> GeomProcessor for GeoJsonWriter<'_, W> {
         self.out.write_all(b"]")?;
         Ok(())
     }
+    fn empty_point(&mut self, idx: usize) -> Result<()> {
+        self.comma(idx)?;
+        self.out
+            .write_all(br#"{"type": "Point", "coordinates": []}"#)?;
+        Ok(())
+    }
     fn point_begin(&mut self, idx: usize) -> Result<()> {
         self.comma(idx)?;
         self.out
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by michaelkirk** [comment]
Could you add a test for this?

**Comment by Oreilles** [comment]
Wkt -> GeoJSON test added. I wanted to add the reverse, but turns out the `geojson` create doesn't accept it:

https://github.com/georust/geojson/blob/4cfbf3a951cff12818e30c269606d46fb4c80486/src/util.rs#L211-L215

**Comment by michaelkirk** [comment]
>  turns out the geojson create doesn't accept [empty points].

TBH I didn't realize geojson could represent empty points, but you've convinced me that it seems like a bug in the geojson crate. I've just opened https://github.com/georust/geojson/issues/231

**Review by michaelkirk — APPROVED** [review]
Thanks, LGTM!

**Comment by Oreilles** [comment]
Ok, so after further inspection it seems like PostGIS is the one deviating from the spec here...

https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.1
```
A position is an array of numbers.  There MUST be two or more
   elements.
```

michaelkirk:
heh i noticed the same thing, but I thought maybe it was just sloppy spec writing, and that the null geometry thing seemed to trumped the position thing.

Probably we should be careful with this one, and get some others to weigh in.

PR #165, opened 2023-08-18
Add GDAL read support for more types

Oreilles:
This PR adds GDAL read support for CircularString, CompoundCurve, CurvePolygon, Triangle, MultiCurve, PolyhedralSurface, TIN, MultiSurface. The implementation is mapped from the Wkb reader.

I initially wanted to also add write support, but didn't manage to do so with the current design. Considering that GDAL write support doesn't handle 3D geometries and GeometryCollection (according to the commented out tests), a bigger refactor might be the opportunity to adress those issues and add support for these new tyeps as well.

This PR also includes a refactor of `fn type2d` to use `gdal_sys::OGR_GT_Flatten(wkb_type)` to get the 2D geometry type. Is it safe to do so ?

### DIFF
--- a/geozero/src/gdal/gdal_reader.rs
+++ b/geozero/src/gdal/gdal_reader.rs
@@ -62,32 +62,61 @@ fn process_geom_n<P: GeomProcessor>(geo: &Geometry, idx: usize, processor: &mut 
             }
             processor.geometrycollection_end(idx)
         }
+        OGRwkbGeometryType::wkbCircularString => process_circularstring(geo, idx, processor),
+        OGRwkbGeometryType::wkbCompoundCurve => process_compoundcurve(geo, idx, processor),
+        OGRwkbGeometryType::wkbCurvePolygon => process_curvepolygon(geo, idx, processor),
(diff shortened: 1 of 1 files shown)


geozero/src/gdal/gdal_reader.rs
```diff
@@ -147,12 +195,71 @@ fn process_polygon<P: GeomProcessor>(
...
+    processor.curvepolygon_begin(ring_count, idx)?;
+    for i in 0..ring_count {
+        let geom = unsafe { geo.get_unowned_geometry(i) };
+        process_curve(&geom, i, processor)?;
+    }
+    processor.curvepolygon_end(idx)
+}
+
 #[cfg(test)]
 #[cfg(feature = "with-wkt")]
```
pka, line 258: We need a feature test for `with-wkb` here.

pka (changes requested):
Looks good for me, with one minor thing, thanks!

geozero/src/gdal/gdal_reader.rs
Oreilles: Done 👍

pka:
Thanks!

PR #166, opened 2023-08-20
Fix WKB and WKT empty points handling

Oreilles:
Currently:
- A WKB geometry `POINT EMPTY` would be written in WKT as `POINT(NaN, NaN)`
- A WKB geometry `MULTIPOINT(1 2,EMPTY,3 4)` would be written in WKT as `MULTIPOINT(1 2,Nan Nan,3 4)`

This PR fixes both issues by calling `processor.empty_point` in the WKB reader if the parsed WKB coordinates are `NaN`, and by writing `EMPTY` instead of `{x} {y} ...` in the WKT writer if the coordinates given are `NaN`.

Notes, cause this seems redundant:
- An empy point could also just be a point with `NaN` coordinates. Assuming that `NaN` coordinates are already supported by all readers (`NaN` is a `f64`), we might be able to get rid of the special `GeoProcessor::empty_point` ?
- A coordinate could actually be an untagged point...
- ...Then the two bugs would become a single one


geozero/src/wkb/wkb_reader.rs
```diff
@@ -667,6 +701,12 @@ mod test {
             //OGR: MULTIPOINT ((10 -20 100),(0 -0.5 101))
         );
 
+        // SELECT 'MULTIPOINT (10 -20 100, 0 -0.5 101)'::geometry
```
michaelkirk, line 704: This comment looks like a copy paste error - can you update?

Oreilles: Fixed - If you notice other small stuff like that, please feel free to commit directly to the branch as I allowed edits by maintainers! Alternatively the "Add suggestion" comment feature is a nice way to solve these in a more seamless way. 🙂

PR #170, opened 2023-08-28
Compile proto file only when vector_tile.rs doesn't exist

pka:
So protoc is not required when used as a dependency.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,7 @@ postgis = "0.9.0"
 postgres = "0.19"
 postgres-types = "0.2"
 prost = "0.11.9"
+# NOTICE: When updating prost-build, remove geozero/src/mvt/vector_tile.rs to force a rebuild
 prost-build = "0.11.9"
 scroll = "0.11"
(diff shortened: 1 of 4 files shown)


nyurik:
this looks ok, but my only concern is that since the output of the generation is inside the `src/` dir, it is not clear to anyone that in order to re-generate it one should delete the file.  Moreover, I am worried that as proto-gen dependency gets updated, we may not realize that the file needs to be regenerated because in the future the generated output might be different... Perhaps at least add a comment next to the proto dependency in Cargo.toml to delete this file?    In general, I am not a big fan of magical dependency skipping

nyurik:
@pka you didn't reply to any of the above concerns - are they moot for some reason?

pka:
Sorry for the silence!

Your concerns should be adressed in ca52c299862ecdcda7d334238390337da8a6f8f4

In the light of the serde precompiled macro drama I came to the conclusion, that build time change of a source file is to dangerous anyway, which made me confident that we should update vector_tile.rs manually only.

PR #172, opened 2023-09-08
geozero-cli: HTTP FGB can write to any output

michaelkirk:
Previously only SVG and json outputs were allowed.

There was a separate code path for handling http input, and we (myself included) failed to keep that code path updated as new output formats were added. This PR combines the code paths.


michaelkirk:
The build failure (new clippy lints from latest rust, unrelated to this PR) addressed in #173

pka (approved):
:+1:

pka:
Remark: With the latest https://github.com/pka/http-range-client we could now read all formats over HTTP (or at least single file formats). But that's another PR.

PR #176, opened 2023-09-29
Include error details

michaelkirk:
Before:
Processing failed: I/O error

After:
Processing failed: I/O error `No such file or directory (os error 2)`

### DIFF
--- a/geozero/src/error.rs
+++ b/geozero/src/error.rs
@@ -37,13 +37,13 @@ pub enum GeozeroError {
     #[error("processing geometry `{0}`")]
     Geometry(String),
     // General
-    #[error("I/O error")]
+    #[error("I/O error `{0}`")]
     IoError(#[from] std::io::Error),
     #[cfg(feature = "with-mvt")]
-    #[error("MVT error")]
+    #[error("MVT error `{0}`")]
     MvtError(#[from] crate::mvt::MvtError),
     #[cfg(feature = "with-gdal")]
-    #[error("GDAL error")]
+    #[error("GDAL error `{0}`")]
     GdalError(#[from] crate::gdal::GdalError),
 }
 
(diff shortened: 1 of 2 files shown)


nyurik (approved):

nyurik:
I think we should start using https://docs.rs/insta/latest/insta/ instead :)  Much much much easier for all sorts of unit/integration testing

michaelkirk:
> So you want to get started with snapshot testing? Here is the 5 minute introduction to how to be successful with insta.

I don't know what insta is, but looking briefly at the website I don't understand what this has to do with this pull request?

nyurik:
@michaelkirk the reason your PR is failing is because there is an assert in the code, validating code response against a hardcoded expected value. `insta` is an amazing system for managing those expected values - so you can either run `cargo test` and it will pass/fail, or you can `cargo insta review` - where you can "bless" new expected values interactively.  The actual expected values are stored in a separate snapshot folder managed by the insta too.

nyurik:
P.S. this was more of a general idea rather than a specific suggestion to this PR

michaelkirk:
Oh I see. I don't think the challenge of copy/pasting a new string was the issue in this case. 

Now, if there is some sort of fancy rust crate that will slap me in the face whenever I forget to set up my per-project git hooks, I will `cargo install` it so fast.

nyurik:
@michaelkirk you are in luck... Create a `.cargo-husky/hooks/pre-push` file and add this to Cargo.toml:

```toml
[dev-dependencies]
cargo-husky = { version = "1", features = ["user-hooks"], default-features = false }
```

An example of it:  https://github.com/nyurik/automotive_diag/blob/main/.cargo-husky/hooks/pre-push

frewsxcv (approved):

pka:
> I think we should start using https://insta.rs/docs/quickstart/ instead :) Much much much easier for all sorts of unit/integration testing

I also think that Insta is a great fit for geometry testing. It came out only a little too late for geozero. Maybe we can start using it when adding a new driver?

PR #177, opened 2023-10-12
forbid unsafe

nyurik:


nyurik:
dope, silly me, missed the unsafe usage. I guess we will have to keep allowing unsafe, unless we figure out a  way not to use it

PR #178, opened 2023-11-07
address latest clippy lints (from rust-1.73)

michaelkirk:
Fixes lint failures like this:

```
warning: unneeded `return` statement with `?` operator
   --> geozero/src/mvt/mvt_reader.rs:261:13
    |
261 |             return Err(MvtError::GeometryFormat)?;
    |             ^^^^^^^ help: remove it
```

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -69,9 +69,3 @@ jobs:
         env:
           DATABASE_URL: "${{ steps.pg.outputs.connection-uri }}?sslmode=disable"
 
-      - name: Ensure valid cfg usage
-        shell: bash
-        run: |
-          rustup toolchain install nightly
-          export RUSTFLAGS='-D warnings'
-          cargo +nightly check -Z unstable-options -Z check-cfg --workspace --all-features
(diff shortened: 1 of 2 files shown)


nyurik (approved):

michaelkirk:
Hm... `cargo +nightly check` is failing against some of our dependencies. 

It might be a legitimate problem or might just be some problem with nightly. 

Do we really want to be running +nightly as part of our CI?

nyurik:
The nightly check was intended for something else -- to catch config errors like [these examples](https://doc.rust-lang.org/nightly/unstable-book/compiler-flags/check-cfg.html#examples). It is only available in nightly. We should probably disable any other linting warnings though to reduce the noise levels

michaelkirk:
I don't think the tradeoff for running `cargo +nightly check` is worth the value of linting typos in our cfgs, so I've proposed removing it: https://github.com/georust/geozero/pull/178/commits/c3f76cf0a6455109b43d0e11522d0164e18c9dc1`

What do other people think?

If someone else (@nyurik?) wants to fix the broken check to be more robust, I'd also be fine to keep it.

frewsxcv:
+1 to not running nightly in CI. Nightly breaks things regularly enough that's it's not worth the tradeoff IMHO

nyurik:
sounds good to me too - config check should probably be commented out until it becomes stable to remind us to re-enable it? Unless it will be part of the warnings by default(?)

PR #179, opened 2023-11-19
update flatgeobuf

michaelkirk:
## Flatgeobuf release notes 4.0.0

https://github.com/flatgeobuf/flatgeobuf/blob/master/src/rust/CHANGELOG.md#400---2023-10-14

> [4.0.0] - 2023-10-14
> 
> Breaking: select_all and select_bbox now return a FeatureIter instead of a modified Self type.
> Ensure reading from tls is supported.
> Breaking: Added flatgeobuf::Error with more specific errors and use it where possible rather than geozero::GeozeroError.
> Added feature batching for HTTP client to optimize remote reading.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ dbase = "0.4"
 diesel = { version = "2.1.0", default-features = false, features = ["postgres"] }
 dup-indexer = "0.3"
 env_logger = "0.10.0"
-flatgeobuf = "3.27.0"
+flatgeobuf = "4.0.0"
 futures-util = "0.3.28"
 gdal = { version = "0.16", default-features = false }
 gdal-sys = "0.9"
--- a/geozero-bench/benches/geobench.rs
+++ b/geozero-bench/benches/geobench.rs
@@ -1,6 +1,6 @@
 use criterion::{criterion_group, criterion_main, Criterion};
-use geozero::error::Result;
 use geozero::ToGeo;
+type Result<T> = std::result::Result<T, Box<dyn std::error::Error>>;
 
 #[derive(Copy, Clone, Debug, PartialEq)]
 pub struct Extent {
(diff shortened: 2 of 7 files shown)


nyurik:
I am a bit confused by the new error format (I'm not saying this is a blocker, just a bit uncertain why this pattern is used):  usually, the more specific errors are wrapped by `thiserror` as `#[error(transparent)] SubError(#[from] ErrorSubType)` -- which allows everyone to use the top level error, and the `?` would auto-convert the sub-error to top-error. Is this pattern no longer good?

michaelkirk:
Can you be more specific about the alternative you’re implying Yuri? In what crate would these wrapped errors be defined? On Nov 20, 2023, at 03:16, Yuri Astrakhan ***@***.***> wrote:﻿
I am a bit confused by the new error format (I'm not saying this is a blocker, just a bit uncertain why this pattern is used):  usually, the more specific errors are wrapped by thiserror as #[error(transparent)] SubError(#[from] ErrorSubType) -- which allows everyone to use the top level error, and the ? would auto-convert the sub-error to top-error. Is this pattern no longer good?

nyurik:
Hmm, I guess this is once again that circular dependency issue.

Usually, I would expect this to be added to the `GeozeroError` enum:

```rust
#[derive(Error, Debug)]
pub enum GeozeroError {
    ...
    #[error(transparent)]
    FlatGeoBufError(#[from] flatgeobuf::Error),
}
```

Possibly with some conditional compilation.  This way any code like this would not require any boxing (as you do in the tests) or the conversion function as in CLI:

```rust
        let ds = HttpFgbReader::open(&args.input).await?
```

michaelkirk:
I'm not sure if that would be a good idea, but since we can't create a circular dependency anyway, I'm not sure it's worth going through the thought exercise.

PR #181, opened 2023-11-27
[WIP] Partial implementation of the MVT writer

nyurik:
Experimenting with MVT writer implementation to support property generation

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ criterion = "0.5.1"
 csv = "1.2.2"
 dbase = "0.4"
 diesel = { version = "2.1.0", default-features = false, features = ["postgres"] }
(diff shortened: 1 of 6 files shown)


michaelkirk:
Hey @nyurik - any particular hold up that caused this to stall out or just the ole' problem of "time and energy is finite"

Any plans to continue with this in the near term?

nyurik:
@michaelkirk well put :)  I would LOVE to work on it, but would need to find time, so if you want to take over that, please do.  It is, sadly, not as trivial as I was hoping.  BTW, totally unrelated, but an important heads-up: MapLibre is actively working with various players (like Microsoft) on the next generation of the tile spec, most likely based on the in-progress work by Markus in https://github.com/mactrem/cov-tiles

Issue #182, opened 2023-12-11
Clarification of `GeomProcessor` methods


kylebarron:
The documentation for the methods in the `GeomProcessor` trait is kinda sparse and that makes it hard to know what order of traits is guaranteed from data producers.

I'm trying to flesh out support in geoarrow for conversion from geozero to geoarrow arrays. Up until now, I've made the user choose a geometry trait when importing the trait, e.g. `ToPolygonArray`, and to some extent assume the inputs will indeed be polygons.

Now I'm trying to implement support from _arbitrary-typed_ or mixed type items (see https://github.com/geoarrow/geoarrow-rs/pull/304), which means storing a piece of internal state for each geometry input so that when `xy()` is called, I know which geometry type that's associated with, and thus which internal geometry array to append to.

But one thing that isn't documented is when each method gets called, when it's a part of a larger geometry. So for example in the GEOS reader, `point_begin` gets called for a point type https://github.com/georust/geozero/blob/8501fedf288adc851d08a367f8f651ee643d4bed/geozero/src/geos/geos_reader.rs#L49 but **not** for a MultiPoint type or any other type. But [`linestring_begin`](https://github.com/georust/geozero/blob/8501fedf288adc851d08a367f8f651ee643d4bed/geozero/src/geos/geos_reader.rs#L147) gets called for all of line string, multi line string, polygon, and multi polygon types! 

I guess this is just a feature request to flesh out these docstrings? It's hard to ensure conformance against the traits only by reading the code.

nyurik:
@kylebarron valid point. I don't have an answer to your question, but a suggestion on how we can all chip in with this - could you start a pull request to improve the doc strings, adding a few of the ones you have figured out, and then the reviewers can make suggestions/changes to that?  This way we can all collaborate on it before merging.

PR #183, opened 2023-12-27
Attempt at trait documentation

kylebarron:
This is relatively a shot in the dark, and I have no idea how much of this is accurate. I'm trying to implement geozero APIs in geoarrow and it's hard to know what I can count on and what I can't. 

Closes https://github.com/georust/geozero/issues/182


nyurik (commented):
Love it, thanks!!!  I think it could mostly be merged as is.  One suggestion that might help the novice users - put more of this info in the trait's overall documentation rather than into individual methods. For example, `FeatureProcessor` doc should explain that its members will be called in the following order (and document the order, with links to the actual methods). Then you don't actually need most of the per-method documentation, apart from parameter documentation.

geozero/src/geometry_processor.rs
```diff
@@ -123,21 +126,46 @@ pub trait GeomProcessor {
...
+    ///
+    /// ## Parameters
+    ///
+    /// - `size`: the number of Points in this MultiPoint
+    /// - `idx`: the positional index of this MultiPoint. This will be 0 except in the case of a
+    ///   GeometryCollection.
+    ///
+    /// ## Following events
+    ///
+    /// - [`point_begin`][`Self::point_begin()`] for each contained point (NOTE: seems expected but seems not implemented)
```
michaelkirk, line 139: I just created https://github.com/georust/geozero/issues/184 with a longer explanation. Can you link to it here in the docs?

geozero/src/feature_processor.rs
```diff
@@ -6,22 +6,46 @@ use crate::property_processor::PropertyProcessor;
...
+    /// ## Invariants
+    ///
+    /// - `dataset_end` is called _only once_ for an entire dataset.
+    /// - No other methods may be called after `dataset_end`.
     fn dataset_end(&mut self) -> Result<()> {
         Ok(())
     }
     /// Begin of feature processing
+    ///
+    /// - `idx` refers to the positional row index in the dataset. For the `n`th row, `idx` will be
```
michaelkirk, line 30:
```suggestion
    /// - `idx` the positional row index in the dataset. For the `n`th row, `idx` will be
```

nit: seems like a couple wasted words (here and duplicated elsewhere)

geozero/src/geometry_processor.rs
```diff
@@ -123,21 +126,46 @@ pub trait GeomProcessor {
...
     ///
-    /// An untagged `LineString` is either a Polygon ring or part of a `MultiLineString`
+    /// ## Parameters
     ///
-    /// Next: size * xy/coordinate
+    /// - `tagged`: if `false`, this `LineString` is either a Polygon ring or part of a `MultiLineString`
+    /// - `size`: the number of coordinates in this LineString
+    /// - `idx`: the positional index of this LineString. This will be 0 for a tagged LineString
+    ///   except in the case of a GeometryCollection. This can be non-zero for an untagged
+    ///   LineString for MultiLineStrings or Polygons with multiple interiors
```
michaelkirk, line 163:
```suggestion
    ///   LineString for MultiLineStrings or Polygons with multiple interiors.
```

geozero/src/geometry_processor.rs
```diff
@@ -186,6 +246,16 @@ pub trait GeomProcessor {
     }
 
     /// Begin of `GeometryCollection` processing
+    ///
+    /// ## Parameters
+    ///
+    /// - `size`: the number of geometries in this GeometryCollection.
+    /// - `idx`: the positional index of this GeometryCollection. Unsure when this is non-zero?? Nested geometry collections?
```
michaelkirk, line 253: GeoZero has FeatureProcessing and GeometryProcessing apis and the interaction between the two can be ill defined at times.

For example, a format like geojson supports the _feature_ processing APIs, allowing you to enumerate each feature's geometry and properties. But geojson also supports the Geometry-only processing apis. So you can iterate over each feature, but only process the geometry - ignoring its properties.

In this case, the `n`th feature of a feature collection will have a geometry with index `n`.
See https://github.com/georust/geozero/blob/main/geozero/src/geojson/geojson_reader.rs#L140

A related issue is https://github.com/georust/geozero/pull/35, wherein processing feature collections as "geometry only" produces invalid output for some formats.

geozero/src/property_processor.rs
```diff
@@ -42,6 +44,18 @@ pub enum ColumnValue<'a> {
 #[allow(unused_variables)]
 pub trait PropertyProcessor {
     /// Process property value. Abort processing, if return value is true.
+    ///
+    /// - `idx` is the positional index of the column??? The row of the dataset? Unknown
```
michaelkirk, line 48: positional index of the property.

```
{
  "geometry": [...],
  "properties": {
     "name": "michael",
     "age": 12
  },
  ...
}
```

name has index 0, age has index 1

michaelkirk (changes requested):

geozero/src/geometry_processor.rs
kylebarron: I removed the mentions of `point_begin` and `point_end` and linked to #184 for discussion

geozero/src/geometry_processor.rs
kylebarron: I'm not sure if I fully understood that. Does that mean that the text I added for `multipolygon_begin` is wrong:

> - `idx`: the positional index of this MultiPolygon. This will be 0 except in the case of a GeometryCollection.

I.e. is that idx the row of the dataset?

michaelkirk: Maybe this example will clarify:

```
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "geometry": {
        "type": "Point",
        "coordinates": [ -18, 44 ]
      },
      "properties": {
        "name": "Point A",
        "description": "This is the first point."
      }
    },
    {
      "type": "Feature",
      "geometry": {
        "type": "Point",
        "coordinates": [ 71, 58 ]
      },
      "properties": {
        "name": "Point B",
        "description": "This is the second point."
      }
    }
  ]
}
```

When you process this with a FeatureProcessor, you'll get something like:
```
begin_dataset
  begin_feature 0 
    begin_geometry 0
    end_geometry 0
  end_feature 0
  begin_feature 1
    begin_geometry 0
    end_geometry 0
  end_feature 1
end_dataset
```

Notice that in the above, the args to `begin_geometry` are always 0.

If you process this with GeometryProcessor, I think you'll get something different:
```
begin_dataset
  begin_geometry 0
  end_geometry 0
  begin_geometry 1
  end_geometry 1
end_dataset
```

Notice the args to `begin_geometry` are _not_ always 0 in this example.

I think both of these behaviors make sense, but it's a little tricky.

kylebarron: I kinda interpret that as "begin_geometry passes the `idx` of the geometry". When passed via `FeatureProcessor`, each row is its own geometry and so the `idx` is always `0`. When passed via `GeometryProcessor`, all rows are considered as one `GeometryCollection`, and so each row has its own `idx`

michaelkirk: That interpretation makes sense to me. So in response to your comment in the code:

> Unsure when this is non-zero?? Nested geometry collections?

1. yes, nested geometry collections
2. but also when you're using GeometryProcessor to process a _Feature_ whose geometry is a GeometryCollection. As in:
```
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "geometry": {
        "type": "GeometryCollection",
        "geometries": [ ... ]
      },
      "properties": {
        "name": "Feature A",
        "description": "This is the first feature."
      }
    },
    {
      "type": "Feature",
      "geometry": {
        "type": "GeometryCollection",
        "geometries": [ ... ]
      },
      "properties": {
        "name": "Feature B",
        "description": "This is the second feature."
      }
    }
  ]
}
```

begin_geometry will be called with a non-zero index when GeometryProcessing gets to "Feature B".

kylebarron: Every time I come back to this PR I'm tired and my mind hurts trying to understand this 🙃 . How do you like the rewording I just added?

michaelkirk: Perfecto. Sorry about the brain pain.

michaelkirk (approved):

michaelkirk:
(As an aside, seems like the merge queue is working 👍)

PR #186, opened 2024-01-07
Remove Arrow mod, point to geoarrow crate

kylebarron:
This is marked as WIP until 0.1 release of the geoarrow crate, expected to be this week (I need arrow-rs to release a new version).

- This removes the `arrow` mod in favor of the https://github.com/geoarrow/geoarrow-rs crate, which implements geozero conversions to and from each of the array types. 
- arrow2 was an alternate implementation of the Arrow spec, but has [stopped being maintained](https://github.com/jorgecarleitao/arrow2/graphs/contributors), so I think the ecosystem (except polars) is focusing on arrow-rs. I don't think there's much benefit in keeping the existing code here.


michaelkirk:
>  which implements geozero conversions to and from each of the array types.

In light of #187, do you think it's desirable to have the geozero integrations live externally? TBH I haven't really figured out which way is "least worst".

kylebarron:
I agree it's a difficult question. My first thought is: the dependency should go "from the crate that is changing more to the crate that is changing less". Right now, `geoarrow` is changing heavily while `geozero` is relatively stable, so it seems like having the geozero implementation in the `geoarrow` crate makes for easier and faster development iteration. 

In relation to #187 and shapefile, maybe it's the other way around; if the core part of parsing shapefiles is more stable than geozero's release cycle, then maybe geozero itself should depend on `geozero-shp`'s public APIs?

Just my initial thoughts; what do you think?

michaelkirk:
> from the crate that is changing more to the crate that is changing less

It's unfortunately not a very objective or stable criteria to apply, but I do think it's a pragmatic approach. 👍 

One small thing to clarify: I don't think we care about non-breaking changes, only semver incompatible changes that can complicate cargo's dependency resolution.

In that sense, the shapefile code might have a better chance of being stable, just by virtue of the format itself being stable. But I suppose that's not necessarily so.

kylebarron:
> It's unfortunately not a very objective or stable criteria to apply, but I do think it's a pragmatic approach. 👍

I agree it's not objective, but it feels like Rust's strict version resolution pushes towards this in practice. 

With geoarrow specifically, the geozero bindings are pretty significant in size. [This entire directory](https://github.com/geoarrow/geoarrow-rs/tree/main/src/io/geozero) is just for geozero bindings. It seems like it would be more invasive to add all of that to geozero and for maintenance to happen here.

kylebarron:
FWIW: I published v0.1 of `geoarrow` and its geozero integration is here: https://docs.rs/geoarrow/0.1.0/geoarrow/io/geozero/index.html

pka:
I'm +1 for merging. If there is an external `arrow` based implementation, we can drop the `arrow2` based implementation.

michaelkirk:
Just to make sure I understand, the change notes for this might read like:

> Moved the geozero/geoarrow integration from the geozero crate to the geoarrow crate

And the motivation for the change is because the geo-arrow crate is undergoing more frequent breaking changes than geozero (currently anyway).

If so, it's all fine with me.

michaelkirk (approved):

kylebarron:
It's my first time trying the merge queue and I'm excited!

michaelkirk:
Thanks for the explanation - I feel like you've explained it before, but it's kind of complicated for someone not steeped in the day to day of arrow, so I appreciate the recap. 👍

PR #188, opened 2024-01-09
Make other wkb and wkt variants generic

kylebarron:
This is a follow up to https://github.com/georust/geozero/pull/171 to make other WKB and WKT variants generic over any `AsRef<[u8]>` input

With this change, I think `WktString` and `WktStr` are redundant, but I didn't want to make a breaking change. I _think_ this should be non-breaking because `String` and `&str` can `AsRef` to `[u8]` right?


kylebarron:
Is the failing CI still a change in this PR? It seems unrelated

michaelkirk:
> Is the failing CI still a change in this PR? It seems unrelated

I think this will be fixed here https://github.com/georust/geozero/pull/189

geozero/src/wkb/wkb_reader.rs
```diff
@@ -26,38 +26,38 @@ impl<B: AsRef<[u8]>> GeozeroGeometry for Wkb<B> {
 )]
 #[cfg_attr(feature = "with-postgis-diesel", diesel(sql_type = Geometry))]
 #[cfg_attr(feature = "with-postgis-diesel", diesel(sql_type = Geography))]
-pub struct Ewkb(pub Vec<u8>);
```
michaelkirk, line 29: What do you think about a default trait implementation to make this change non-breaking (same for GpkgWkb, SpatiaLiteWkb, and MySQLWkb)?

geozero/src/wkt/wkt_reader.rs
```diff
@@ -7,34 +7,34 @@ use wkt::Geometry;
 
 /// WKT String.
 #[derive(Debug)]
-pub struct WktString(pub String);
+pub struct WktString<B: AsRef<[u8]>>(pub B);
```
michaelkirk, line 10: I agree that it's nice to be able to take either a String or a &str, but I think it's confusing to explain why we have two different functionally equivalent types like this.

What would you think about unifying them? Either by deprecating one or (my preference) revert and deprecate both legacy types and introduce a third unified struct (maybe called `Wkt`?).

(same for Ewkt)

michaelkirk (changes requested):

geozero/src/wkt/wkt_reader.rs
kylebarron: I agree that having multiple redundant types is confusing; I wasn't sure which would be preferred, but I'm happy to deprecate these and move to `Wkt`

> I agree that it's nice to be able to take either a String or a &str

It's also nice that it can take in a `Vec<u8>` and `&[u8]` at the same time! There were a couple places that previously had copies with `to_vec` that could be removed

geozero/src/wkb/wkb_reader.rs
kylebarron: Oh I didn't realize that this was breaking, but I suppose it's breaking because the internal type is `pub`? So it still _accepts input_ with `Vec<u8>` but when you call `Ewkb(...).0` you no longer know you're getting a `Vec<u8>`?

kylebarron: > default trait implementation

Do you have an example of what you mean?

kylebarron:
I haven't worked with deprecated objects in rust before and I don't know the right way to get clippy happy with deprecated but not yet removed objects. Thoughts?

geozero/src/wkb/wkb_reader.rs
michaelkirk: Like this: `pub struct Ewkb<B: AsRef<[u8]> = Vec<u8>>(pub B);`

michaelkirk:
> I haven't worked with deprecated objects in rust before and I don't know the right way to get clippy happy with deprecated but not yet removed objects. Thoughts?

I would slap some `#[allow(deprecated)]` on our own internal usage.

geozero/src/wkb/wkb_reader.rs
kylebarron: Interesting!

kylebarron:
```
error: defaults for type parameters are only allowed in `struct`, `enum`, `type`, or `trait` definitions
  --> geozero/src/wkb/wkb_reader.rs:29:17
   |
29 | pub struct Ewkb<B: AsRef<[u8]> = Vec<u8>>(pub B);
   |                 ^^^^^^^^^^^^^^^^^^^^^^^^
```

did I do that wrong?

michaelkirk: Hm. No, you did it right, but it looks like a bug in diesel's derive macros:

Here's an annotated snippet from cargo expand:

```
impl<'expr, B: AsRef<[u8]> = Vec<u8>> AsExpression<Geometry> for &'expr Ewkb<B> {
                          ^^^^^^^^^^- THIS IS A PROBLEM!!!
    type Expression = Bound<Geometry, Self>;
    fn as_expression(self) -> Self::Expression {
        Bound::new(self)
    }
}
```

I'll file a bug report with the diesel people. I guess we should just eat the breaking change?

michaelkirk: Filed: https://github.com/diesel-rs/diesel/issues/3902

kylebarron: 🤷‍♂️

michaelkirk:
Rather than wait for diesel to make a release, I say we merge this and accept that it's a small breaking change to our diesel integration (we're already looking at a semver breaking release as it is).

michaelkirk:
> I say we merge this 

Sorry to be sloppy with my language, I mean to say that we should revert the default trait implementation that I earlier suggested, since it's breaking the current release of diesel.

And *then* merge this PR.

kylebarron:
Do you want to keep the deprecated structs instead of removing them in this release?

michaelkirk:
> Do you want to keep the deprecated structs instead of removing them in this release?

Even though this is already going to be a breaking release, I'd prefer to keep them. As a user, I know I like having deprecations that tell me exactly what needs to change.

kylebarron:
Awesome, any other comments on this PR?

michaelkirk (approved):

ariesdevil:
When do we plan to release the new version on crates.io?

PR #190, opened 2024-01-10
expose feature for gdal with bindgen

michaelkirk:
This is necessary for linking against gdal versions whose bindings
aren't yet included in gdal-sys, which is common when using a newly
released gdal.

Also removed redundant gdal-sys dependency - gdal already includes that,
and geozero doesn't use it directly.

~~edit: note this won't pass CI until #189 is merged~~ **update**: rebased!

Issue #184, opened 2023-12-30
`point_begin`/`point_end` should be called around `xy` for each Point in a MultiPoint


michaelkirk:
Extracted from https://github.com/georust/geozero/pull/183

When processing MultiPoints, many of the formats that implement GeometryProcessor do not call `point_begin`/`point_end` for each point in a multipoint, and instead call only `xy`. To be consistent with MultiLineString and MultiPolygon, we should surround each child geometry with it's corresponding begin/ends.

In truth, I don't know of a format yet that *requires* these events, so I expect most implementations will be a no-op, but I think making the usage consistent across `Multi{Point,LineString,Polygon}` makes things easier to understand.

```diff
processor.begin_multipoint(...)
for point in multipoint {
+  processor.begin_point(...)
   processor.xy(...)
+  processor.end_point(...)
}
processor.end_multipoint(...)
```

kylebarron:
I don't think this is possible without a breaking change because `point_begin` does not have a `tagged` argument.

PR #191, opened 2024-01-10
Release v0.12.0?

michaelkirk:
I've drafted up some release notes.

We still have a couple nearly-merged PR's that would be good to include.

#188, #189, #190

What do people think?


kylebarron:
I'm happy to get more of my PRs resolved before the 0.12 release, but I need some more feedback. I don't know what the appetite for breaking changes is, but if there isn't much appetite for breaking changes, then https://github.com/georust/geozero/issues/184 can be closed and https://github.com/georust/geozero/pull/183 might be able to be finished

michaelkirk:
> I don't know what the appetite for breaking changes is, but if there isn't much appetite for breaking changes, then https://github.com/georust/geozero/issues/184 can be closed

I hit the same confusion: https://github.com/michaelkirk/geomedea/blob/main/geomedea_geozero/src/geozero_reader.rs#L92

I think I'm in favor of the change - in my experience, the existing behavior is surprising and leads to bugs. Part of making that change would be updating all the formats in this repo and ideally notifying any others that we know of.

That said, there's no reason I know of that it has to happen in this release.

re: #183 - I'll review shortly.

kylebarron:
> I hit the same confusion: https://github.com/michaelkirk/geomedea/blob/main/geomedea_geozero/src/geozero_reader.rs#L92

😅 

> I think I'm in favor of the change - in my experience, the existing behavior is surprising and leads to bugs. Part of making that change would be updating all the formats in this repo and ideally notifying any others that we know of.
> 
> That said, there's no reason I know of that it has to happen in this release.

I agree that I'm in favor of the change, agree that it's a bigger lift because you have to ensure writers are up to date, and agree it doesn't need to happen in this release.

kylebarron:
Once we have #186 merged, I'd love to get an 0.12 release out! Let me know if I can help with this at all

kylebarron:
I created #196 with an updated changelog from some of my recent PRs. Feel free to merge those changes back in here if you'd prefer.

michaelkirk:
superseded by #196

PR #193, opened 2024-01-17
geojson lines writer

kylebarron:
This works for features but not geometries because it's hard to know when to insert the `\n` to end the line.


geozero/src/geojson/geojson_line_writer.rs
```diff
@@ -0,0 +1,190 @@
...
+    #[ignore = "fails because no newlines are written. feature_end is never called."]
+    #[test]
+    fn good_geometries() {
+        let input = r#"{ "type": "Point", "coordinates": [1.1, 1.2] }
+    { "type": "Point", "coordinates": [2.1, 2.2] }
+    { "type": "Point", "coordinates": [3.1, 3.2] }
+    "#;
+        let mut out: Vec<u8> = Vec::new();
+        assert!(
+            read_geojson_lines(input.as_bytes(), &mut GeoJsonLineWriter::new(&mut out)).is_ok()
```
michaelkirk, line 161: Maybe we could count how many `*_begin` are called, and then we could print a newline whenever the corresponding number of `*_end` are called.

As long as the data and processor is valid, that should work. Though maybe #184 precludes that from working with points. If we can't catch this with the compiler, maybe a runtime error is better than quietly producing invalid output?

michaelkirk: Oh I forgot #184 only affects multipoint... I think something like this might work:
```
GeoJsonLineWriter {
   open_contexts: 0
   ...
}

impl GeoJsonLineWriter {
    fn begin_context(&mut self) {
        self.open_contexts +=1;
    }
    fn end_context(&mut self) {
        self.open_context -=1;
        if self.open_contexts == 0 {
            self.write_newline();
        }
    }
} 

impl FeatureProcessor for GeoJsonLineWriter {
    fn feature_begin(&mut self) {
       self.begin_context()
    }
    fn feature_end(&mut self) {
       self.end_context()
    }
}
impl GeomProcessor for GeoJsonLineWriter {
    fn point_begin(&mut self) {
        self.begin_context()
        ...
    }
    fn point_end(&mut self) {
        self.end_context()
        ...
    }
    fn linestring_begin(&mut self) {
        self.begin_context()
        ...
    }
    fn linestring_end(&mut self) {
        self.end_context()
        ...
    }

}
```

kylebarron: That makes sense, thanks for the idea!

michaelkirk (changes requested):
Whoops, sorry that this sat here so long. I forgot about it. 

Feel free to merge after addressing that last small change.

geozero/src/geojson/geojson_line_writer.rs
```diff
@@ -0,0 +1,227 @@
...
+            self.write_newline()?;
+        }
+        Ok(())
+    }
+}
+
+impl<W: Write> FeatureProcessor for GeoJsonLineWriter<W> {
+    fn feature_begin(&mut self, _idx: u64) -> Result<()> {
+        self.begin_context();
+        self.line_writer.feature_begin(0)?;
```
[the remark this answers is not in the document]
kylebarron: I'll have to look back at this but I thought that it had to be 0. Otherwise it added an extra comma or something like that to the feature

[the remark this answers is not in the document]
michaelkirk: Ah! That could be. That's worth a comment then.

geozero/src/geojson/geojson_line_writer.rs
```diff
@@ -0,0 +1,227 @@
...
+        self.line_writer.property(idx, name, value)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::geojson::read_geojson_lines;
+
+    #[ignore = "fails because no newlines are written. feature_end is never called."]
```
kylebarron, line 190: Hmm I need to check this ignore statement as well

kylebarron:
It turns out that passing `idx` instead of `0` fails the test for the _feature_, but it also doesn't pass the geometry test. For example, when logging to stdout I see:

```
failures:

---- geojson::geojson_line_writer::tests::good_geometries stdout ----
a: {"type": "Point", "coordinates": [1.1,1.2]}
b: { "type": "Point", "coordinates": [1.1, 1.2] }
a: ,{"type": "Point", "coordinates": [2.1,2.2]}
b: { "type": "Point", "coordinates": [2.1, 2.2] }
thread 'geojson::geojson_line_writer::tests::good_geometries' panicked at geozero/src/geojson/geojson_line_writer.rs:227:73:
called `Result::unwrap()` on an `Err` value: Error("expected value", line: 1, column: 1)
```
Because the idx passed into `begin_point` is >0, a comma gets added here https://github.com/georust/geozero/blob/458f38ccca37f0ac4341593491017220b805ed51/geozero/src/geojson/geojson_writer.rs#L22-L27

kylebarron:
I found a workaround and added a comment. Let me know if that solution is ok

Issue #199, opened 2024-03-01
Write WKT with precision


kylebarron:
It doesn't appear that there's currently a way to write WKT values with a given precision? 

I'm implementing "display methods" in geoarrow so that Python users can have a better display in a terminal than
```
<geoarrow.rust.core._rust.MultiPolygonArray at 0x10a12e9a0>
```
and in this use case, having too many decimals of precision is distracting. E.g. shapely gives a "repr" of 

<img width="664" alt="image" src="https://github.com/georust/geozero/assets/15164633/728c52ac-0fa1-4bd5-8985-396659c9d24d">

It looks like this `format!` would have to be changed?
https://github.com/georust/geozero/blob/52a4d2d3c11f02e734274fcb6ee4b88b94b5b53d/geozero/src/wkt/wkt_writer.rs#L104

Ref https://stackoverflow.com/a/49778870

PR #200, opened 2024-03-01
Pass formatter parameters to WKT writer

kylebarron:
This is just a prototype of #199. 

Based on the example [here](https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.precision), this optionally passes down a `precision` parameter to the `format!()` call.

One thing I noticed when I hard-coded `format!("{x:.3} {y:.3}")` is that setting that forces _even integers_ to have 3 decimal points. E.g. 
```
---- geojson::geojson_reader::test::conversions stdout ----
thread 'geojson::geojson_reader::test::conversions' panicked at geozero/src/geojson/geojson_reader.rs:419:9:
assertion `left == right` failed
  left: "POINT(10.000 20.000)"
 right: "POINT(10 20)"
```

Potentially this means what I actually want to do to solve my goal in #199 is to truncate the precision of my _data_, and that will automatically get printed as I want...

### DIFF
--- a/geozero/src/wkt/mod.rs
+++ b/geozero/src/wkt/mod.rs
@@ -26,20 +26,21 @@ pub(crate) mod conversion {
             dialect: WktDialect,
             dims: CoordDimensions,
             srid: Option<i32>,
+            precision: Option<usize>,
         ) -> Result<String>;
     }
 
     impl<T: GeozeroGeometry> ToWkt for T {
         fn to_wkt(&self) -> Result<String> {
-            self.to_wkt_with_opts(WktDialect::Wkt, CoordDimensions::default(), None)
+            self.to_wkt_with_opts(WktDialect::Wkt, CoordDimensions::default(), None, None)
         }
 
         fn to_ewkt(&self, srid: Option<i32>) -> Result<String> {
-            self.to_wkt_with_opts(WktDialect::Ewkt, CoordDimensions::xyzm(), srid)
+            self.to_wkt_with_opts(WktDialect::Ewkt, CoordDimensions::xyzm(), srid, None)
         }
 
         fn to_wkt_ndim(&self, dims: CoordDimensions) -> Result<String> {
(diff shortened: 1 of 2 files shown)


michaelkirk:
> truncate the precision of my data, and that will automatically get printed as I want...

I think that approach will work (1.0 - 0.07) == 0.9299999999999999... percent of the time. 😉

michaelkirk:
Dumb jokes aside, you mean something like this, right?

```
let y = (x * 10.pow(n)).round() / 10.pow(n);
format!("{}", y);
```

I think that should work 100 percent of the time, though opens us up to loss of precision at high values. 

I don't know why, but I'm sort of surprised that a formatting option like this doesn't already exist.

kylebarron:
> Dumb jokes aside, you mean something like this, right?
> 
> ```
> let y = (x * 10.pow(n)).round() / 10.pow(n);
> format!("{}", y);
> ```

Yeah. That's how I ended up implementing it in geoarrow: https://github.com/geoarrow/geoarrow-rs/blob/667389b2a1cdae704b815b6a447505027d876ff9/src/io/display/scalar.rs#L18-L21

> opens us up to loss of precision at high values.

Hmm yeah that's true. I'm not sure how to truncate floats with high precision.

> I don't know why, but I'm sort of surprised that a formatting option like this doesn't already exist.

You mean in core rust? Maybe it does; I haven't searched for it.

frewsxcv:
```rust
fn main() {
    let num = 3.14159;
    let decimals = 3;
    let formatted = format!("{:.1$}", num, decimals);
    println!("{}", formatted);  // Output: 3.142
}
```

kylebarron:
@frewsxcv that's what I'm doing here:
https://github.com/georust/geozero/pull/200/files#diff-553c2186acf21c8c360979841037ea9343c11e4e2937438cb9eca2c20d0d08cdR107-R109

But we'd still need a way for the user to pass down the precision desired

Issue #202, opened 2024-03-06
`geos::Geometry` to `ewkt`  panic


lewiszlw:
```
    #[test]
    fn geos_to_ewkt() {
        use geozero::GeozeroGeometry;
        use geozero::ToWkt;

        let wkt = "POINT(1 1)";
        let mut ggeom = geos::Geometry::new_from_wkt(wkt).unwrap();

        println!("ewkt: {}", ggeom.to_ewkt(ggeom.srid()).unwrap());
    }
```
this test will panic with 
```
thread 'function::as_ewkt::tests::geos_to_ewkt' panicked at /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/geos-8.3.1/src/coord_seq.rs:498:9:
assertion failed: self.nb_dimensions >= 3
```

PR #203, opened 2024-03-06
Fix `geos:Geometry` to `ewkt` bug

lewiszlw:
Fixes https://github.com/georust/geozero/issues/202.

image

### DIFF
--- a/geozero/src/geos/geos_reader.rs
+++ b/geozero/src/geos/geos_reader.rs
@@ -101,12 +101,17 @@ fn process_coord_seq<P: GeomProcessor>(
 ) -> Result<()> {
     let multi = processor.dimensions().z;
     let n_coords = cs.size()?;
+    let cs_dims = cs.dimensions()?;
     for i in 0..n_coords {
         if multi {
             processor.coordinate(
                 cs.get_x(i)?,
                 cs.get_y(i)?,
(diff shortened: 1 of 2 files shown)


geozero/src/geos/geos_reader.rs
```diff
@@ -101,12 +101,17 @@ fn process_coord_seq<P: GeomProcessor>(
...
     let n_coords = cs.size()?;
+    let cs_dims = cs.dimensions()?;
     for i in 0..n_coords {
         if multi {
             processor.coordinate(
                 cs.get_x(i)?,
                 cs.get_y(i)?,
-                Some(cs.get_z(i)?),
+                if matches!(cs_dims, geos::CoordDimensions::ThreeD) {
+                    Some(cs.get_z(i)?)
```
pka, line 111: Just out of curiosity - why doesn't a `cs_dims == geos::CoordDimensions::ThreeD` work?

pka (changes requested):
Thanks fo the fix!

Could you please add your test case as well?

geozero/src/geos/geos_reader.rs
lewiszlw: `==` works here. Changed.

pka (approved):
👍

Issue #204, opened 2024-03-16
Trait error using sqlx example


nk9:
I'm using the [SQLX geo-types example](https://github.com/georust/geozero#postgis-usage-examples), and it's failing. Has the required code changed? 

```rust
use geo::{Geometry, Point};

let geom: Geometry<f64> = Point::new(10.0, 20.0).into();
let _ = sqlx::query(
    "INSERT INTO point2d (datetimefield,geom) VALUES(now(),ST_SetSRID($1,4326))",
)
.bind(wkb::Encode(geom))
.execute(&pool)
.await?;
```

```
 1  error[E0277]: the trait bound `geozero::wkb::Encode<geo::Geometry>: sqlx::Encode<'_, _>` is not satisfied
    --> rust/load_all.rs:245:19
     |
 245 |             .bind(wkb::Encode(geom))
     |              ---- ^^^^^^^^^^^^^^^^^ the trait `sqlx::Encode<'_, _>` is not implemented for `geozero::wkb::Encode<geo::Geometry>`
     |              |
     |              required by a bound introduced by this call
     |
     = help: the following other types implement trait `sqlx::Encode<'q, DB>`:
               <bool as sqlx::Encode<'q, sqlx::Any>>
               <bool as sqlx::Encode<'_, Postgres>>
               <i8 as sqlx::Encode<'_, Postgres>>
               <i16 as sqlx::Encode<'q, sqlx::Any>>
               <i16 as sqlx::Encode<'_, Postgres>>
               <i32 as sqlx::Encode<'q, sqlx::Any>>
               <i32 as sqlx::Encode<'_, Postgres>>
               <i64 as sqlx::Encode<'q, sqlx::Any>>
             and 34 others
 note: required by a bound in `Query::<'q, DB, <DB as HasArguments<'q>>::Arguments>::bind`
    --> /Users/nick/.cargo/registry/src/index.crates.io-6f17d22bba15001f/sqlx-core-0.7.3/src/query.rs:79:32
     |
 79  |     pub fn bind<T: 'q + Send + Encode<'q, DB> + Type<DB>>(mut self, value: T) -> Self {
     |                                ^^^^^^^^^^^^^^ required by this bound in `Query::<'q, DB, <DB as HasArguments<'q>>::Arguments>::bind`
```

nk9:
Ah, this works if you install the required feature:

```
cargo add geozero --features with-postgis-sqlx
```

[...]

PR #205, opened 2024-03-16
Add feature requirement info to README and docs

nk9:
I believe these are correct, and are in all the relevant locations. But please confirm!

This addresses #204 in the simplest way possible.


kylebarron (approved):

nyurik (approved):
lgtm, thx!

PR #208, opened 2024-04-12
Handle nested properties in GeoJSON reader and writer

kylebarron:
The GeoJSONReader currently silently omits any nested property. This serializes the `Value` for an Array or Object back to a string to store within a `ColumnValue::Json`.

The GeoJSONWriter currently silently omits any `ColumnValue::Json` or `ColumnValue::Binary`. This PR writes the contained JSON string to the output writer. It leaves writing binary to a future PR (which should probably at least have the option of converting to hex?)

I'm not sure why but I was unable to run tests locally because they failed with 
```
  = note: ld: library 'pq' not found
```
I'm guessing this has something to do with Postgres' libpq, but `brew install libpq` says it's already installed.

### DIFF
--- a/geozero/src/geojson/geojson_reader.rs
+++ b/geozero/src/geojson/geojson_reader.rs
@@ -1,4 +1,4 @@
-use crate::error::Result;
+use crate::error::{GeozeroError, Result};
 use crate::{
     ColumnValue, FeatureProcessor, GeomProcessor, GeozeroDatasource, GeozeroGeometry,
     PropertyProcessor,
@@ -217,7 +217,17 @@ pub(crate) fn process_properties<P: PropertyProcessor>(
                 processor.property(i, key, &ColumnValue::ULong(v.as_u64().unwrap()))?
             }
             JsonValue::Bool(v) => processor.property(i, key, &ColumnValue::Bool(*v))?,
-            // Null, Array(Vec<Value>), Object(Map<String, Value>)
+            JsonValue::Array(v) => {
+                let json_string =
+                    serde_json::to_string(v).map_err(|_err| GeozeroError::Property(key.clone()))?;
+                processor.property(i, key, &ColumnValue::Json(&json_string))?
+            }
+            JsonValue::Object(v) => {
+                let json_string =
+                    serde_json::to_string(v).map_err(|_err| GeozeroError::Property(key.clone()))?;
+                processor.property(i, key, &ColumnValue::Json(&json_string))?
+            }
+            // Null
             _ => processor.property(i, key, &ColumnValue::String(&value.to_string()))?,
         };
     }
--- a/geozero/src/geojson/geojson_writer.rs
+++ b/geozero/src/geojson/geojson_writer.rs
@@ -445,7 +445,7 @@ mod test {
     }
 
     #[test]
-    fn nested_property() {
+    fn nested_object_property() {
         let geojson = r#"{
   "type": "FeatureCollection",
   "features": [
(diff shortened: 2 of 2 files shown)


geozero/src/geojson/geojson_writer.rs
```diff
@@ -208,6 +208,12 @@ fn write_str_prop<W: Write>(mut out: W, colname: &str, v: &str) -> Result<()> {
     Ok(())
 }
 
+fn write_json_prop<W: Write>(mut out: W, colname: &str, v: &str) -> Result<()> {
+    let colname = colname.replace('\"', "\\\"");
```
michaelkirk, line 212: What's this escaping about? Is this captured in a test somewhere?

kylebarron: That's identical to the other `write*` methods
https://github.com/georust/geozero/blob/cdda9a8da2136b2bbda7517561906969296f9081/geozero/src/geojson/geojson_writer.rs#L199
https://github.com/georust/geozero/blob/cdda9a8da2136b2bbda7517561906969296f9081/geozero/src/geojson/geojson_writer.rs#L205

Any embedded `"` in the column name must be escaped with a `\"` or else the JSON will be invalid. I'm not sure if there's an existing test for this, but it would be separate from this change at hand.

geozero/src/geojson/geojson_reader.rs
```diff
@@ -207,20 +207,30 @@ pub(crate) fn process_properties<P: PropertyProcessor>(
...
+            JsonValue::Number(v) => {
+                if v.is_f64() {
+                    processor.property(i, key, &ColumnValue::Double(v.as_f64().unwrap()))?
+                } else if v.is_i64() {
+                    processor.property(i, key, &ColumnValue::Long(v.as_i64().unwrap()))?
+                } else if v.is_u64() {
+                    processor.property(i, key, &ColumnValue::ULong(v.as_u64().unwrap()))?
+                } else {
+                    unreachable!()
+                }
```
kylebarron, line 219: This was refactored so that the outer `match value` could be exhaustive.

geozero/src/geojson/geojson_writer.rs
michaelkirk: Ok, I created https://github.com/georust/geozero/issues/209 as a followup task.

kylebarron:
Anything else I can do to push this along?

michaelkirk (approved):
LGTM

geozero/src/geojson/geojson_reader.rs
```diff
@@ -207,20 +207,30 @@ pub(crate) fn process_properties<P: PropertyProcessor>(
...
+            JsonValue::Object(v) => {
+                let json_string =
+                    serde_json::to_string(v).map_err(|_err| GeozeroError::Property(key.clone()))?;
+                processor.property(i, key, &ColumnValue::Json(&json_string))?
             }
-            JsonValue::Bool(v) => processor.property(i, key, &ColumnValue::Bool(*v))?,
             // For null values omit the property
             JsonValue::Null => false,
-            // Array(Vec<Value>), Object(Map<String, Value>)
-            _ => processor.property(i, key, &ColumnValue::String(&value.to_string()))?,
```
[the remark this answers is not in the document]
kylebarron: Hmm maybe I conflated the reading and writing parts in the PR description. I don't remember exactly right now. 

I think the primary change is that we change from the `String` variant to `Json`. I don't know what was the behavior of `value.to_string()`. I assume that was not guaranteed to be the same as `serde_json::to_string`?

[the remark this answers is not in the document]
michaelkirk: I think the output is the same. I still think your change to the Reader makes sense - if we have JSON we might as well put it in a `ColumnValue::JSON` rather than a `ColumnValue::String`. I just wanted to get the changelog notes correct.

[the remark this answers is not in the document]
michaelkirk: see changelog entry at https://github.com/georust/geozero/pull/213/files#diff-f088ff93ac6d7d6d266d38c61f69ae356b4fc823196a9fd6e991f93eed1376b7R13

PR #210, opened 2024-04-24
Ewkt support srid=0

kkk25641463:
I think it's nessary  to support srid=0 with ewkt format.
SRID of 0 doesn't technically exist, it just means no SRID or maintain format consistency

See [also](https://docs.snowflake.com/en/sql-reference/functions/st_aswkb)

### DIFF
--- a/geozero/src/wkt/wkt_writer.rs
+++ b/geozero/src/wkt/wkt_writer.rs
@@ -45,7 +45,7 @@ impl<W: Write> WktWriter<W> {
         if self.first_header && self.dialect == WktDialect::Ewkt {
             self.first_header = false;
             match srid {
-                None | Some(0) => (),
+                None => (),
                 Some(srid) => self.out.write_all(format!("SRID={srid};").as_bytes())?,
             }
         }


michaelkirk:
I'm not really familiar with this. Is it possible to provide a test that show an example of the value of this change?

nyurik:
hmm, technically this is correct - no SRID and SRID=0 are the same... I don't see a problem, but we do need a test case

michaelkirk:
What I don't understand is, if they are the same, what's the problem with the existing approach that writes `SRID=0`?

nyurik:
useless info?  We don't want to have two different cases for identical state, so ideally `srid=nnn` should only be used for non-zero case?

kkk25641463:
Below is the test case. I don't know which test case can satisfy more scenarios.

PostGis:
```
SQL:  SELECT ST_ASEWKT('SRID=0;LINESTRING(0.75 0.75, -10 20)')
OUTPUT: LINESTRING(0.75 0.75,-10 20)
```
[SnowFlake](https://docs.snowflake.com/en/sql-reference/functions/st_asewkt):
```
SQL: SELECT ST_ASEWKT('SRID=0;LINESTRING(0.75 0.75, -10 20)')
OUTPUT:  SRID=0;LINESTRING(0.75 0.75,-10 20) 
```

nyurik:
So I am a bit confused now - why do we want to explicitly differentiate between `None` and `SRID=0`?  I would prefer to treat them the same and not print `SRID=0` at all...   Unless of course `EWKT` requires the SRID to always be set, in which case we must output `None` as `SRID=0`

nyurik:
It seems EWKT supports all the standard WKT cases, e.g. `SELECT ST_GeomFromEWKT('POINT(-71.064544 42.28787)');` works just fine - so it is really up to us.  I guess for EWKT case we should do the EWKT-way of always specifying SRID:

```rust
        if self.first_header && self.dialect == WktDialect::Ewkt {
            self.first_header = false;
            self.out.write_all(format!("SRID={};", srid.unwrap_or_default()).as_bytes())?
        }
```

kkk25641463:
Understood, thank you for your analysis.

Issue #211, opened 2024-05-01
Unresolved import `gdal_sys`


alecsandrei:
When trying to compile geozero with gdal I am getting an error.

```sh
[dependencies]
geozero = { version = "0.12.0", features = ["with-gdal"] }
```

```sh
error[E0432]: unresolved import `gdal_sys`
 --> /home/alex/.cargo/registry/src/index.crates.io-6f17d22bba15001f/geozero-0.12.0/src/gdal/gdal_error.rs:2:5
  |
2 | use gdal_sys::OGRwkbGeometryType;
  |     ^^^^^^^^ use of undeclared crate or module `gdal_sys`

error[E0432]: unresolved import `gdal_sys`
 --> /home/alex/.cargo/registry/src/index.crates.io-6f17d22bba15001f/geozero-0.12.0/src/gdal/gdal_reader.rs:4:16
  |
4 | use gdal_sys::{self, OGRwkbGeometryType};
  |                ^^^^ no external crate `gdal_sys`

error[E0432]: unresolved import `gdal_sys`
 --> /home/alex/.cargo/registry/src/index.crates.io-6f17d22bba15001f/geozero-0.12.0/src/gdal/gdal_writer.rs:6:5
  |
6 | use gdal_sys::OGRwkbGeometryType;
  |     ^^^^^^^^ use of undeclared crate or module `gdal_sys`
```
With version 0.11.0 it works fine.

kylebarron:
Are you able to share a minimal reproducible example? E.g. a tiny public repo with only geozero showing what line of code created this error?

alecsandrei:
Sure, it's just an empty project with the dependency specified in Cargo.toml.

https://github.com/alecsandrei/geozero_test

kylebarron:
It looks like it works if you add `gdal-sys` as a feature:

```
geozero = { version = "0.12.0", features = ["gdal-sys", "with-gdal-bindgen"] }
```
that works with your git repo for me

PR #212, opened 2024-05-13
make `with-gdal` feature require `gdal-sys`

kylebarron:
`gdal-sys` is its own optional dependency, and currently `with-gdal` does not require it. That means that any import of `gdal-sys` from geozero will fail without the user turning on an explicit `gdal-sys` feature, see #211. So e.g. this fails:

https://github.com/georust/geozero/blob/8f3d8e4b9363aee0d50356936f5be65f496b9953/geozero/src/gdal/gdal_error.rs#L2

I don't see why `with-gdal` wouldn't turn on `gdal-sys`.

Closes https://github.com/georust/geozero/issues/211

### DIFF
--- a/geozero/Cargo.toml
+++ b/geozero/Cargo.toml
@@ -14,7 +14,7 @@ license.workspace = true
 [features]
 default = ["with-svg", "with-wkt", "with-geo", "with-geojson"]
 with-csv = ["dep:csv", "with-wkt"]
-with-gdal = ["dep:gdal"]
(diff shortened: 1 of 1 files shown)


michaelkirk (approved):
I broke this in #190.

> Also removed redundant gdal-sys dependency - gdal already includes that, and geozero doesn't use it directly.

I must have typo'd whatever grep convinced me that we weren't using gdal_sys directly. Sorry.

kylebarron:
Oh oops. Well I did approve it 🙂

PR #214, opened 2024-05-14
Convert capabilities into tabular format

frewsxcv:


michaelkirk (approved):
Nice!

michaelkirk:
Oh there are a few more formats following your table - were those intentionally excluded?

frewsxcv:
> Oh there are a few more formats following your table - were those intentionally excluded?

Not intentionally excluded. Just pushed a commit with the rest of the formats.

Issue #138, opened 2023-04-09
Error refactoring


nyurik:
GeoZero has one `Error` enum, and many of the values are `String` objects that contain the actual error information, sometimes formatted.

I think it would better to refactor it so that each specific error can be easily analyzed without any extra performance cost.  

### Proposed structure

```rust
enum GeozeroError {
    arrow: ArrowError, csv: CsvError, gdal: GdalError, geojson: GeojsonError, geos: GeosError, gpkg: GpkgError, gpx: GpxError, mvt: MvtError, postgis: PostgisError, svg: SvgError, tessellator: TessellatorError, wkb: WkbError, wkt: WktError,
}
```

Each one of the sub-errors are their own enums with the actual values. So instead of this in `src/mvt/mvt_reader.rs`:

```rust
return Err(GeozeroError::Feature(format!(
    "invalid feature.tags length: {:?}", feature.tags.len()
)));
```

we will have this:

```rust
return Err(MvtError::InvalidFeatureTagsLength(feature.tags.len());
```

Note that [thiserror](https://docs.rs/thiserror/latest/thiserror/) has `#[from]` to simplify `MvtError -> GeozeroError` conversion.

### Existing errors
Here are all the error usages at the moment (some of these are used multiple times)

```
GeozeroError::ColumnNotFound
GeozeroError::ColumnType(stringify!($e), format!("{v:?}")
GeozeroError::Dataset(error
GeozeroError::Feature("invalid feature.tags length: {feature_tags_count:?}"
GeozeroError::Feature("invalid key index {key_idx}"
GeozeroError::Feature("invalid value index {value_idx}"
GeozeroError::Geometry("CoordSeq missing"
GeozeroError::Geometry("Invalid UTF-8 encoding"
GeozeroError::Geometry("Missing Geometry"
GeozeroError::Geometry("Missing LineStrings for Polygon"
GeozeroError::Geometry("Missing container for LineString"
GeozeroError::Geometry("Missing container for Polygon"
GeozeroError::Geometry("Missing polygons for MultiPolygon"
GeozeroError::Geometry("No LineStrings for MultiLineString"
GeozeroError::Geometry("No coords for LineString"
GeozeroError::Geometry("No coords for MultiPoint"
GeozeroError::Geometry("No coords for Point"
GeozeroError::Geometry("Not ready for coords"
GeozeroError::Geometry("The input was an empty Point, but the output doesn't support empty Points"
GeozeroError::Geometry("Too few coordinates in line or ring"
[...]

PR #216, opened 2024-05-28
CSV Error Refactor

nrhill1:
Hey @nyurik ,

I made an error module for `csv` in response to #138 . Let me know if you think any changes need to be made.


geozero/src/csv/csv_error.rs
```diff
@@ -0,0 +1,14 @@
+//! CSV Error type.
```
nyurik, line 1: i usually just add the doc string to `CsvError` - doing it module level seem a bit weird tbh, esp because the module is not public - only the enum is.

nyurik (commented):
looks good - i just would move the doc string a bit, and no objections otherwise

geozero/src/csv/csv_error.rs
```diff
@@ -0,0 +1,14 @@
...
+use thiserror::Error;
```
nrhill1, line 2: hmmmm I am slightly confused since isn't that the case for `MvtError` as well?

geozero/src/csv/csv_error.rs
```diff
@@ -0,0 +1,14 @@
...
+
+#[derive(Error, Debug)]
+pub enum CsvError {
+    #[error("column not found or null")]
+    ColumnNotFound,
+    #[error("Invalid UTF-8 encoding")]
+    InvalidUtf8,
+    #[error("error processing dataset: `{0}`")]
+    Processing(String),
```
michaelkirk, line 11: This seems unused?

geozero/src/csv/csv_error.rs
```diff
@@ -0,0 +1,14 @@
+//! CSV Error type.
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum CsvError {
+    #[error("column not found or null")]
+    ColumnNotFound,
+    #[error("Invalid UTF-8 encoding")]
+    InvalidUtf8,
```
michaelkirk, line 9: This seems unused?

geozero/src/csv/csv_error.rs
```diff
@@ -0,0 +1,14 @@
...
+    #[error("error processing dataset: `{0}`")]
+    Processing(String),
+    #[error("error parsing to WKT `{0}`")]
+    WktError(String),
```
michaelkirk, line 13: #138 wasn't my request, but my understanding is that the point was to move from `String` errors to structured errors, but this is still a String error.

nrhill1: I'm using a `String` buffer here in order to display the error message because `Wkt` from the `wkt` crate uses a `&'static str` for errors.

michaelkirk: Ah, so then making it more structured is out of your hands.  

But there's no value in copying the &'static str to a String is there? In that case, it's preferable to `WktError(&'static str)` since it'll avoid duplicating the contents of the &str onto the heap.

geozero/src/csv/csv_error.rs
nrhill1: Hi Michael,

I wanted to use the `InvalidUtf8` variant in `csv/mod.rs` to replace `GeozeroError::Geometry`, however `CsvError` will not convert to `GeozeroError` to match the custom `Result` type. I created the `Processing` variant to replace `GeozeroError::Dataset` on line 217 of `csv_reader.rs`, but I'm also running into a similar problem there.

geozero/src/csv/csv_error.rs
```diff
@@ -0,0 +1,10 @@
+//! CSV Error type.
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum CsvError {
+    #[error("column not found or null")]
```
michaelkirk, line 6: My understanding of @nyurik's request in #138 was to avoid heap allocations.


That doesn't apply to GeozeroError::ColumnNotFound - so I don't see the benefit of adding this CsvError::ColumnNotFound. 

Is there some benefit I'm missing? If not, please revert that portion of the PR and continue to use GeozeroError::ColumnNotFound.

geozero/src/csv/csv_reader.rs
```diff
@@ -109,17 +111,14 @@ pub fn process_csv_geom(
...
 
     let mut collection_started = false;
 
     for (record_idx, record) in reader.into_records().enumerate() {
         let record = record?;
-        let geometry_field = record
-            .get(geometry_idx)
-            .ok_or(GeozeroError::ColumnNotFound)?;
-        let wkt = wkt::Wkt::from_str(geometry_field)
-            .map_err(|e| GeozeroError::Geometry(e.to_string()))?;
```
michaelkirk, line 122: Here you've gotten read of a to_string (heap allocation), so good job, so I think this one makes sense.

geozero/src/csv/csv_error.rs
michaelkirk: Please move the doc to the enum. It's not a big deal, but no sense in doubling down on "wrong" just for consistency.

michaelkirk (changes requested):
Sorry this sat here so long @nrhill1

geozero/src/csv/csv_error.rs
michaelkirk: I think #138 has two goals, having an error enum per format and making sure those error enums avoid String allocations. 

I don't have strong feelings about the former, and this is a pretty inoffensive change. So in the interest of moving forward, I'm going to merge this as is (after moving the doc comment).

michaelkirk (approved):

michaelkirk (approved):

michaelkirk:
Merged with `main` and fixed some clippy errors.

PR #220, opened 2024-07-17
Avoid copying string for WKT parsing

kylebarron:
Currently bytes are converted into a `Read` and then collected _back_ into a string. It would be preferable to pass `[u8]` or `str` directly into `wkt::Wkt::from_str`

### DIFF
--- a/geozero/src/wkt/wkt_reader.rs
+++ b/geozero/src/wkt/wkt_reader.rs
@@ -2,6 +2,7 @@ use crate::error::{GeozeroError, Result};
 use crate::{FeatureProcessor, GeomProcessor, GeozeroDatasource, GeozeroGeometry};
 
 use std::io::Read;
+use std::str::FromStr;
 use wkt::types::{Coord, LineString, Polygon};
 use wkt::Geometry;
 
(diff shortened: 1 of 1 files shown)


geozero/src/wkt/wkt_reader.rs
```diff
@@ -11,7 +12,10 @@ pub struct Wkt<B: AsRef<[u8]>>(pub B);
 
 impl<B: AsRef<[u8]>> GeozeroGeometry for Wkt<B> {
     fn process_geom<P: GeomProcessor>(&self, processor: &mut P) -> Result<()> {
-        read_wkt(&mut self.0.as_ref(), processor)
+        let wkt_str = std::str::from_utf8(self.0.as_ref())
```
kylebarron, line 15: I know we've been changing the API repeatedly lately, but it would be nice here if the user could declare they already have UTF8 data. I.e. maybe `WktStr<S>` where `S: AsRef<str>` separated from `Wkt<B>` where `B: AsRef<[u8]>`

michaelkirk: > I know we've been changing the API repeatedly lately, but it would be nice here if the user could declare they already have UTF8 data

To make sure I understand, you're proposing that, we remove the `WktStr` deprecation and make it generic? Sounds reasonable, just making sure I understand.

michaelkirk (approved):

geozero/src/wkt/wkt_reader.rs
kylebarron: Yes, I believe so. Just so that the buffers don't have to be checked again via `std::str::from_utf8`

PR #224, opened 2024-07-26
Upgrade SQLX from 0.7 to 0.8

### LINKED ISSUE #223: Upgrade SQLX from 0.7 to 0.8
There is a new version of SQLX.

sergiomeneses:
closes #223

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,10 +7,7 @@ resolver = "2"
 [workspace.package]
 # This version should match the version in [workspace.dependencies] below
 version = "0.13.0"
-authors = [
-    "Pirmin Kalberer <pka@sourcepole.ch>",
-    "Yuri Astrakhan <YuriAstrakhan@gmail.com>",
-]
+authors = ["Pirmin Kalberer <pka@sourcepole.ch>", "Yuri Astrakhan <YuriAstrakhan@gmail.com>"]
 edition = "2021"
 homepage = "https://github.com/georust/geozero"
 repository = "https://github.com/georust/geozero"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -27,7 +30,9 @@ clap = { version = "4.3", features = ["derive"] }
 criterion = "0.5.1"
 csv = "1.2.2"
 dbase = "0.4"
-diesel = { version = "2.1.0", default-features = false, features = ["postgres"] }
+diesel = { version = "2.1.0", default-features = false, features = [
```
nyurik, line 33: @sergiomeneses please disable or adjust your toml auto-formatter - while i usually like those, the newlines it introduces here and above are not very common in rust -- especially in the dependencies list.

sergiomeneses: Done.

geozero/src/postgis/postgis_sqlx.rs
```diff
@@ -144,7 +144,8 @@ macro_rules! impl_sqlx_postgis_encode {
             fn encode_by_ref(
                 &self,
                 buf: &mut sqlx::postgres::PgArgumentBuffer,
-            ) -> sqlx::encode::IsNull {
+            ) -> std::result::Result<IsNull, Box<(dyn std::error::Error + Send + Sync + 'static)>>
```
nyurik, line 147: so my understanding here is that sqlx 0.8 modified their Encode trait to handle boxed errors? If so, seems reasonable

sergiomeneses: I update this to qualified use

sergiomeneses:
fix workflow errors.

sergiomeneses:
I think is ready for review.

nyurik (approved):
@sergiomeneses looks good, thanks! I will merge it now with a minor separate fix.  For the future - when you create a PR, make sure to create a branch first.  This way maintainers can make changes in that branch before merging.

PR #228, opened 2024-08-15
Remove this test - it's tripping up CI and redundant anyway.

michaelkirk:
Clippy CI explodes because we're using a non-existent feature. It fails
to compile without DATABASE_URL

```
error: unexpected `cfg` condition value: `dont-compile`
   --> geozero/tests/postgis.rs:258:11
    |
258 |     #[cfg(feature = "dont-compile")]
    |           ^^^^^^^^^^^^^^^^^^^^^^^^
```

But what is this test testing anyway? We already have tests inserting
geo-points into sqlx (see rust_geo_query), so my understanding is that
the thing this test covers is that the sqlx macro behaves the same way,
as the programmatic API, which I'd argue is sqlx's job, not ours.

### DIFF
--- a/geozero/tests/postgis.rs
+++ b/geozero/tests/postgis.rs
@@ -255,7 +255,7 @@ mod postgis_sqlx {
     #[tokio::test]
     #[ignore]
     // Requires DATABASE_URL at compile time
-    #[cfg(feature = "dont-compile")]
+    #[cfg(not(test))] // Delete this line to compile the test
     async fn rust_geo_macro_query() -> Result<(), sqlx::Error> {
         let pool = pg::get_pool().await;
 


geozero/tests/postgis.rs
```diff
@@ -252,63 +252,6 @@ mod postgis_sqlx {
         Ok(())
     }
 
-    #[tokio::test]
-    #[ignore]
-    // Requires DATABASE_URL at compile time
-    #[cfg(feature = "dont-compile")]
```
kylebarron, line 258: Maybe switch to 
```
#[ignore = "requires DATABASE_URL at compile time"]
```

kylebarron:
No opinions on whether the test should remain or not, but I don't think you need the cfg anyways. Or is the issue that `#[ignore]` would still compile the test?

michaelkirk:
> Or is the issue that #[ignore] would still compile the test?

Right.

For some reason I wasn't able to get the `#[allow(unexpected_cfgs)]` to apply. Maybe it's a bad interaction with the tokio test macro? I'm not sure and don't care to dig deeper at this point.

In the meanwhile, I've restored the test, but come up with a way that evades clippy's wrath (for now).

michaelkirk:
From https://doc.rust-lang.org/reference/attributes/testing.html#the-ignore-attribute

> A function annotated with the test attribute can also be annotated with the ignore attribute. The ignore attribute tells the test harness to not execute that function as a test. It will still be compiled when in test mode.

It makes sense that this is supported, but I wish the naming were consistent with the doc tests - e.g. `ignore` (does not build) vs `norun` (builds but doesn't run).

PR #233, opened 2024-08-28
Integrate shp driver

pka:
Solves #187

PR #234, opened 2024-08-28
geozero-shp release with deprecation message

pka:
Should be released & merged before #233

### DIFF
--- a/geozero-shp/Cargo.toml
+++ b/geozero-shp/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "geozero-shp"
-version = "0.5.0"
-description = "Shapefile reader with GeoZero API."
+version = "0.6.0"
+description = "WARNING: The shp driver is now included in geozero."
 readme = "README.md"
 keywords = ["geo", "shapefile"]
 categories = ["science::geo"]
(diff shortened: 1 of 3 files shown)


kylebarron (approved):

michaelkirk (changes requested):
Can you include a module level declaration so that people can see it in their compiler output, since people often don't read documentation:

Something like:
```
#![deprecated(note = "The geozero_shp crate has been deprecated. Instead, use geozero's built-in shapefile feature")]
```

geozero-shp/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "geozero-shp"
-version = "0.5.0"
```
michaelkirk, line 3: I think this should be a patch bump - we're not breaking anything right?

pka:
Test fails because of clippy complaints... Will force merge it when both PRs are ready.

michaelkirk (approved):

pka:
Note to self: Release to crates.io just after next geozero release!

michaelkirk:
Will we have to pin to a specific (old) version of geozero for this geozero_shp release? Currently it's using "workspace"

PR #235, opened 2024-08-29
Drop flatgeobuf dependency in geozero crate

pka:
Solves https://github.com/flatgeobuf/flatgeobuf/issues/283


geozero/tests/geozero-api.rs
```diff
@@ -60,22 +57,16 @@ impl GeomProcessor for MaxHeightFinder {
 }
 
 #[test]
-#[ignore]
 fn max_height_finder() -> Result<()> {
-    let mut filein = BufReader::new(File::open(
-        "tests/data/geoz_lod1_gebaeude_max_3d_extract.fgb",
```
michaelkirk, line 66: I guess this file can also be deleted?

geozero/tests/svg.rs
```diff
@@ -23,6 +19,12 @@ fn json_to_svg() -> Result<()> {
     Ok(())
 }
 
+/*
```
michaelkirk, line 22: If you want to keep this, can you leave a comment for us-in-the-future explaining why it's here but commented out?

geozero/tests/geozero-api.rs
```diff
@@ -86,20 +77,25 @@ impl PropertyProcessor for FeatureFinder {
...
-        let found = feature.process_properties(&mut finder);
-        if found.is_err() || found.unwrap() {
-            break;
-        }
-    }
-    let feature = fgb.cur_feature();
-    let props = feature.properties()?;
-    assert_eq!(props["id"], "DNK".to_string());
-    assert_eq!(props["name"], "Denmark".to_string());
+    // process_properties is not public. We should have iterators!
```
michaelkirk, line 84: Can you explain more what you mean by "we should have iterators"?

My guess is that you mean something like:

```
GeoJsonLineReader::next(&mut self) -> Result<Option<geojson::Feature>> 
```

So that we can write:

```
while let Some(feature) = json.next()? {
    ...
}
```

Is that right? Or something else? If so, that seems reasonable.

---

Though, TBH, this example doesn't seem like a great fit for the geozero API anyway. In my experience, geozero is mostly concerned with processing entire datasets. This example relies mostly on things intrinsic to flatgeobuf, including some API's added to FGB as helpers for implementing geozero APIs, but are not actually geozero APIs themselves.

Cargo.toml
```diff
@@ -43,7 +43,6 @@ hex = "0.4"
 kdbush = "0.2"
 log = "0.4.19"
 lyon = "1.0.1"
-polylabel = "2.5"
```
michaelkirk, line 46: I'm guessing you've removed this example because it uses FGB. We could convert it to use geojson, but I'm guessing you've judged it to be not worth the upkeep?

That seems fine, just making sure I understand.

michaelkirk (commented):
Thanks for taking on this long standing nuisance! 

I'm on board in theory, but I had some questions before proceeding.

geozero/tests/geozero-api.rs
pka: It's on my machine only ;-)

geozero/tests/geozero-api.rs
pka: Exactly. We should have a common feature iterator for `GeozeroDatasource`.

Cargo.toml
pka: Exactly. This example should go either into the flatgeobuf or the polylabel crate. I tried to use geojson, but I had  the same problem with the missing feature iterator.

geozero/tests/geozero-api.rs
kylebarron: I'd really love to have a feature iterator. I've thought briefly about it before but I got stuck because I don't know how to combine the push-based nature of the `GeomProcessor` with the pull-based nature of the iterators

michaelkirk (approved):

michaelkirk:
Everything's looking good, but I'll let you take care of sequencing your various merges @pka.

PR #237, opened 2024-09-03
Fix writing WKT with z/m dimensions

kylebarron:


geozero/src/wkt/mod.rs
```diff
@@ -35,7 +35,7 @@ pub(crate) mod conversion {
         }
 
         fn to_ewkt(&self, srid: Option<i32>) -> Result<String> {
-            self.to_wkt_with_opts(WktDialect::Ewkt, CoordDimensions::xyzm(), srid)
```
kylebarron, line 38: I'm not sure the right way to fix this, but as I understand it, setting `CoordDimensions` says "this output will have these given geometries", and so defaulting to `xyzm` seems incorrect.

geozero/src/geojson/geojson_reader.rs
```diff
@@ -321,11 +321,11 @@ mod test {
         let mut out = WktWriter::with_dims(&mut wkt_data, CoordDimensions::xyz());
         assert!(read_geojson_geom(&mut geojson.as_bytes(), &mut out).is_ok());
         let wkt = std::str::from_utf8(&wkt_data).unwrap();
-        assert_eq!(wkt, "LINESTRING(1 1 10,2 2 20)");
+        assert_eq!(wkt, "LINESTRING Z(1 1 10,2 2 20)");
 
         let geojson = r#"{"type": "LineString", "coordinates": [[1,1],[2,2]]}"#;
         let mut wkt_data: Vec<u8> = Vec::new();
-        let mut out = WktWriter::with_dims(&mut wkt_data, CoordDimensions::xyz());
+        let mut out = WktWriter::with_dims(&mut wkt_data, CoordDimensions::xy());
```
kylebarron, line 328: I'm not sure what this is supposed to test. Is it supposed to test that setting `xyz` doesn't matter? I thought at first below was a typo and that the `LineString` was meant to be 3d, but this entire test is otherwise the exact same as the one before it.

geozero/src/wkt/wkt_writer.rs
```diff
@@ -255,8 +273,18 @@ mod test {
     #[cfg(feature = "with-wkb")]
     fn from_wkb() {
         let blob = hex::decode("01040000A0E6100000020000000101000080000000000000244000000000000034C0000000000000594001010000800000000000000000000000000000E0BF0000000000405940").unwrap();
-        let mut cursor = std::io::Cursor::new(blob);
-        let ewkt = Ewkt::from_wkb(&mut cursor, WkbDialect::Ewkb).unwrap();
-        assert_eq!(ewkt.0, "SRID=4326;MULTIPOINT(10 -20 100,0 -0.5 101)")
+        let ewkt = Ewkb(blob)
+            .to_wkt_with_opts(WktDialect::Ewkt, CoordDimensions::xyz(), Some(4326))
+            .unwrap();
+        assert_eq!(ewkt, "SRID=4326;MULTIPOINT Z(10 -20 100,0 -0.5 101)")
```
kylebarron, line 279: Similarly, the API just isn't sufficient... There's no way to get the coordinate dimension from the `EWKB` without consuming it. And so there's no way right now to know the SRID and CoordDimensions of the input and pass that on to the writer at the beginning.

michaelkirk (dismissed):
These changes seem reasonable, but I didn't write the original code, so I'm not sure of the original motivations.

michaelkirk:
That failing test looks significant:

```
left: "POLYHEDRALSURFACE Z( Z( Z(0 0 0,0 0 1,0 1 1,0 1 0,0 0 0)), Z( Z(0 0 0,0 1 0,1 1 0,1 0 0,0 0 0)), Z( Z(0 0 0,1 0 0,1 0 1,0 0 1,0 0 0)), Z( Z(1 1 0,1 1 1,1 0 1,1 0 0,1 1 0)), Z( Z(0 1 0,0 1 1,1 1 1,1 1 0,0 1 0)), Z( Z(0 0 1,1 0 1,1 1 1,0 1 1,0 0 1)))"
```

kylebarron:
My initial reaction is that it's really hard to fix ZM handling in isolation because it needs some bigger architectural changes to geozero

michaelkirk:
> My initial reaction is that it's really hard to fix ZM handling in isolation

I don't doubt it! It leaves me not sure what to do about this PR though.

PR #238, opened 2024-09-19
Release geozero 0.14.0

pka:

PR #239, opened 2024-09-25
Update Cargo.toml

Dylan-DPC:
Bumps the dependency to at least 2.2.3 as per the advisory

[Advisory](https://rustsec.org/advisories/RUSTSEC-2024-0365.html)

(even though this crate may not be using the function mentioned in the advisory, it is better to bump it to ensure that there is no chance of other dependencies pulling it)


kylebarron (approved):

pka:
Happy to include that in #238, but I'm stuck and hoping for new ideas.

pka:
Merged with #238, thanks!

Issue #245, opened 2024-12-17
Getting geometry and properties from shp file features


nk9:
I have a shapefile. I want to iterate through all of its features with the properties for each. I am going to create a bunch of structs in memory using the data and run operations on them, like getting the area of the geometry and examining properties of each feature. So I need the feature geometries in a format that will work with `geo`'s `unsigned_area` method.

The `ShpReader::iter_features()` function [seems like the one I want](https://docs.rs/geozero/latest/geozero/shp/reader/struct.ShpReader.html#method.iter_features):

> Returns an iterator over the Shapes and their Records

I have examined the [shp tests](https://github.com/georust/geozero/blob/771609474a57e19fde9fdfc03454f1b04613e93d/geozero/tests/shp-reader.rs) to figure out how to use it.

This works for the properties:

```rust
use geozero::shp::ShpReader;
use geozero::{FeatureProperties, ProcessorSink};

fn main() -> Result<(), geozero::shp::Error> {
    let reader = ShpReader::from_path("/path/to/my_shapefile.shp")?;

    for feat in reader.iter_features(&mut ProcessorSink::new())? {
        if let Ok(feat) = feat {
            let props = feat.properties()?;
            println!("{props:?}");
        }
    }

    Ok(())
}
```

But I don't see any way to get the geometries (the `Shape`?) from `feat`. The only field in the `ShapeRecord` is `record` (a `dbase::Record`, which seems to be a key-value data structure), and [the only functions in associated traits](https://docs.rs/geozero/latest/geozero/shp/reader/struct.ShapeRecord.html#trait-implementations) have to do with properties.

Do I actually have to use `iter_geometries` as well? Am I missing something here? I don't see any tests or documentation on how to access geometries from the `iter_features()` function of a `ShpReader`.

kylebarron:
It looks like `iter_features` should automatically pass through the geometries as well: https://github.com/georust/geozero/blob/771609474a57e19fde9fdfc03454f1b04613e93d/geozero/tests/shp-reader.rs#L54

michaelkirk:
You could also consider using the shapefile-rs crate directly, rather than geozero: 

https://github.com/tmontaigu/shapefile-rs/tree/master/examples/geotype-example

https://docs.rs/shapefile/latest/shapefile/reader/index.html

In particular, note the `geo-types` feature, which allows you to get the same structs that the `geo` crate uses, so you can call `unsigned_area` (etc.)

nk9:
[...]

PR #246, opened 2024-12-19
Add `to_geo_feature` to get properties HashMap with geo::Geometry

michaelkirk:
Related to https://github.com/georust/geozero/issues/245 and another comment by @frewsxcv on discord, it seems there's some interest in parsing various data sources to `geo::Geometry` + properties. Here's my attempt!

This works for any FeatureProcessor — like GeoJson, Flatgeobuf, Shapefile, etc.


frewsxcv:
Regarding that last commit, it's because the debug representation changed in https://github.com/georust/geo/pull/1332

PR #251, opened 2025-01-15
CLI adds bounds calculation and style flag to csv writer

michaelkirk:
Adds the ability to specify a top level `<style>` block in your svg output. 

Also exposes this functionality (and the recent bound computation #250) to the CLI's svg output.

It's very basic. So basic that I'm not sure it's worth it, but I'm personally using it so I thought I'd see if it had wider appeal.


pka (approved):
:+1:

PR #252, opened 2025-02-06
chore: upgrade geos

xhwhis:
`geos@9` is yanked. Can you update a new version to crates.io?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,4 +56,4 @@ tokio = { version = "1.30.0", default-features = false }
 wkt = "0.12.0"
 
 [patch.crates-io]
-#geozero = { path = "./geozero" }
+geozero = { path = "./geozero" }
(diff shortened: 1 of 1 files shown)


kylebarron (approved):

michaelkirk:
I pushed up another commit to fix the build by having workspace crates use the local geozero. Let's see how it goes.

/cc @nyurik - you might have feelings about this. Tagging you so you don't think I'm trying to "sneak" it past you. I'm open to other suggestions which solve the problems discussed in my commit.

urschrei:
Very much in agreement with the commit message suggestion FWIW. It's not clear to me why one would want to refer to the published version in this scenario, and it makes maintenance more complex for a small team (which includes people who could not comprehend the RELEASING instructions, i.e. me)

Cargo.toml
```diff
@@ -56,4 +56,4 @@ tokio = { version = "1.30.0", default-features = false }
 wkt = "0.12.0"
 
 [patch.crates-io]
-#geozero = { path = "./geozero" }
```
pka, line 59: git blames me for changing this line in e3f6502634daa2bbf472e56fe26f6137931fa80a. I'm with you that a local build should always use ./geozero. :+1: +1

pka (approved):

PR #257, opened 2025-05-04
Publicize new feature writer structure, so callers can access.

frewsxcv:
---

### DIFF
--- a/geozero/src/geo_types/geo_types_feature_writer.rs
+++ b/geozero/src/geo_types/geo_types_feature_writer.rs
@@ -10,8 +10,8 @@ pub type GeoProperties = HashMap<String, OwnedColumnValue>;
 
 #[derive(Debug, PartialEq)]
 pub struct GeoFeature {
-    pub geometry: geo_types::Geometry,
-    pub properties: GeoProperties,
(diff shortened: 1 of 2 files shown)


frewsxcv:
@michaelkirk I incorporated the feature reading functionality in rgis (https://github.com/frewsxcv/rgis/commit/5142788856f3fb207f74d1cf3d41e2a7b2ce1489) and it works great! Thanks again for adding it

michaelkirk:
Can you explain why you want to do this given the existing implementation already includes `properties`, `properties_mut`, `geometry`, `geometry_mut`, and `into_inner`?

frewsxcv:
> Can you explain why you want to do this given the existing implementation already includes `properties`, `properties_mut`, `geometry`, `geometry_mut`, and `into_inner`?

I didn't see them! Updated the PR to just include the reexport change

michaelkirk:
> I didn't see them! 

Oh phew. I know some people just don't like the ceremony of getter/setter methods, so I'm relieved to not have to slog through that discussion. 😅

michaelkirk (approved):

Issue #254, opened 2025-02-28
Make `build.rs` sandbox-friendly by avoiding writes to the source directory


no-materials:
Hey,

I'm packaging a Rust crate via Nix that depends on geozero and have run into an issue related to its build script when the "with-mvt" feature is enabled. Currently, build.rs forcefully overrides the OUT_DIR environment variable to "src/mvt", which causes the build script to write generated files (e.g., vector_tile.rs) directly into the source tree. In sandboxed and hermetic build environments (such as Nix builds), the source directory is read-only, leading to a "Permission denied" error.

Based on the guidance in the Crane FAQ on sandbox-unfriendly build scripts (https://crane.dev/faq/sandbox-unfriendly-build-scripts.html), build scripts should avoid writing into the source tree and instead use the provided (and writable) OUT_DIR to store generated files. This change would improve compatibility with many CI/CD and package build systems.

Would you consider modifying the build.rs to either:
- Not override OUT_DIR (and instead use Cargo’s provided OUT_DIR), or
- Provide a Cargo feature or environment flag that disables this behavior?

This would make geozero more sandbox-friendly and easier to package in hermetic environments.

Thanks for your attention and all your work on geozero!

PR #259, opened 2025-05-24
Don't overwrite OUT_DIR in build.rs

pka:
Adressing #254

Since the Protobuf generation function in `build.rs` is only called when `vector_tile.rs` doesn't exists, it is effectively never called in a regular build. I'd like to leave `build.rs` in the source tree and at least `OUT_DIR` is not overwritten anymore.

### DIFF
--- a/geozero/build.rs
+++ b/geozero/build.rs
@@ -1,6 +1,5 @@
 #[cfg(feature = "with-mvt")]
 use std::{
-    env,
     fs::OpenOptions,
     io::{Read, Write},
     path::Path,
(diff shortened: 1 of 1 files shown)


nyurik:
one thing i could never understand is why we would ever want to hack OUT_DIR to output something that is clearly temporary to be checked in. Compiling a protobuf file, esp as simple and well studied as the one used for mvt should not result in a checked-in file - but rather use a standard out_dir on every build, conditional on the .proto file not changing

PR #260, opened 2025-06-12
chore: bump dependencies

nyurik:
Update many dependencies, and fix up the code to match the changes.

---


nyurik:
@michaelkirk i also bumped edition to 2024 because now dependencies won't compile without support for that, so makes sense to migrate as well.

michaelkirk:
Oh yeah, I didn't see the 2024 update. 

In `georust/geo` we have a minimum 6 months support window, but I don't feel strongly about requiring that here, so I'm OK with merging it if @pka is.

nyurik:
@michaelkirk I did not update it initially, but there was no point to keep it back because multiple non-optional dependencies were requiring it - so it's basically either that or we would have to postpone this update :shrug:

PR #263, opened 2025-08-29
mvt: expose Command, CommandInteger, ParameterInteger from the module

styrowolf:
---

This PR exposes `mvt::mvt_commands::{Command, CommandInteger, ParameterInteger` as a part of the `mvt` module.

I'm currently working on [a PR to generate MVTs from GeoJSON files for the Martin tile server](https://github.com/maplibre/martin/pull/2098). Since writing features (geom + properties) to MVTs is not supported by `geozero` yet, it could be useful to expose `mvt_commands::{Command, CommandInteger, ParameterInteger` such that users can implement the functionality on their own.

cc: @nyurik

### DIFF
--- a/geozero/src/mvt/mod.rs
+++ b/geozero/src/mvt/mod.rs
@@ -1,5 +1,6 @@
 //! MVT conversions.
 mod mvt_commands;
+pub use mvt_commands::{Command, CommandInteger, ParameterInteger};
 pub(crate) mod mvt_reader;
 pub(crate) mod mvt_writer;
 


styrowolf:
> Perhaps it would be cleaner to implement feature writing for MVT in geozero than publicly exporting internals of MVT generation?

Yeah, that's true. It would be better for the ecosystem. I will start implementing feature writing when I find the time. I'm closing this PR.

> I'm also curious how the mvt module in geozero compares to external mvt crates, such as [mvt](https://docs.rs/mvt/latest/mvt/). (I haven't used geozero for MVT encoding)

The encoding part seems pretty similar on both but geozero's module needs a bit more tooling around it (or feature writing) to build out useful mvt layers.

PR #264, opened 2025-09-01
partial reimplementation of MvtWriter to support feature writing and outputting layers

styrowolf:
---

This is an implementation of feature writing for MVTs, as discussed in #263. This PR builds upon @nyurik's previous WIP reimplementation #181. It passes all of the `mvt_writer` tests. I am not sure if I understood the API exactly correctly but this seems to be a pretty good starting point, as individual feature/geom can be converted to a MVT feature with the `ToMvt` trait as well as `MvtWriter` can process multiple feature and output them to a layer.

[Here is an example](https://github.com/maplibre/martin/blob/20ff5e36cc89bb07aefaf39a6f513df6b65a38cc/martin/src/geojson/mod.rs#L124) of feature writing and layer output with MvtWriter (from GeoJSON to MVT tiles PR, https://github.com/maplibre/martin/pull/2098) 

Let me know how we can move forward with this PR.

### DIFF
--- a/geozero/src/mvt/mvt_writer.rs
+++ b/geozero/src/mvt/mvt_writer.rs
@@ -44,7 +44,7 @@ use super::mvt_error::MvtError;
 ///     })
 ///     .to_string(),
 /// ); // implements GeozeroDatasource
-/// let mut mvt_writer = MvtWriter::new_unscaled(4096);
+/// let mut mvt_writer = MvtWriter::new_unscaled(4096).unwrap();
 /// geojson.process(&mut mvt_writer);
 /// let mvt_layer = mvt_writer.layer("sample"); // returns MVT layer with all features
 /// ```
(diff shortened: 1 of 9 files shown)


styrowolf:
@urschrei the checks fail on something unrelated to the PR. when I run `cargo clippy --workspace --all-features --bins --examples --tests --benches -- -D warnings` on my machine, it passes. do I need to do anything?

michaelkirk:
It's kind of annoying when new clippy lints break builds, but I guess it's a forcing function to improve (usually they're improvements anyway 😆) the codebase.

If you want to speed things along, you can fix the clippy lints in a separate PR.

styrowolf:
Could someone review this PR when possible, so that we can move forward? Appreciate the help.

geozero/src/mvt/mvt_writer.rs
```diff
@@ -24,6 +29,27 @@ pub struct MvtWriter {
     last_y: i32,
     line_state: LineState,
     is_multiline: bool,
+    tags_builder: TagsBuilder,
+}
+
+impl Default for MvtWriter {
```
kylebarron, line 35: Does it make sense to implement `Default` here? The arguments to `new` (`extent: u32, left: f64, bottom: f64, right: f64, top: f64`) don't to me seem to have a natural default.

geozero/src/mvt/tile_value.rs
```diff
@@ -72,6 +73,38 @@ impl TryFrom<Value> for TileValue {
...
+            ColumnValue::Int(v) => TileValue::Sint(*v as i64),
+            ColumnValue::UInt(v) => TileValue::Uint(*v as u64),
+            ColumnValue::Long(v) => TileValue::Sint(*v),
+            ColumnValue::ULong(v) => TileValue::Uint(*v),
+            ColumnValue::Float(v) => TileValue::Float(*v),
+            ColumnValue::Double(v) => TileValue::Double(*v),
+            ColumnValue::String(v) => TileValue::Str(v.to_string()),
+            ColumnValue::Json(v) => TileValue::Str(v.to_string()),
+            ColumnValue::DateTime(v) => TileValue::Str(v.to_string()),
+            ColumnValue::Binary(_) => Err(())?,
```
kylebarron, line 103: Should we base64-encode this?

geozero/src/mvt/mvt_writer.rs
styrowolf: The `default` call creates an "unscaled" MvtWriter, where the geometry is assumed to be in tile coordinates and accepted as is. The `new` call has arguments to scale geometries. Maybe we could rename this.

styrowolf:
Current questions

- Should we encode `ColumnValue::Binary` as base64 string in `TileValue`? https://github.com/georust/geozero/pull/264#discussion_r2322485925
- Are the changes in `MvtWriter` and its implementation of `Default` ready to go? Is there any concerns I've not addressed or something that should be changed?
- Any docs, examples, explanations to add?

Thank you for your time and your reviews.

kylebarron:
> * Should we encode `ColumnValue::Binary` as base64 string in `TileValue`? [partial reimplementation of MvtWriter to support feature writing and outputting layers #264 (comment)](https://github.com/georust/geozero/pull/264#discussion_r2322485925)

I think there should at least be an _option_ to do so. It's probably the most straightforward way of encoding Binary properties, right?

> * Are the changes in `MvtWriter` and its implementation of `Default` ready to go? Is there any concerns I've not addressed or something that should be changed?

I think it would be clearer if it were a separate constructor with an associated docstring. To me, `Default` should be painfully obvious what will happen, and it's not in this case. 

> * Any docs, examples, explanations to add?

I think having a runnable and doctested example would be ideal here. It can be small

styrowolf:
I've added the doc-tested example as well. Ready for (hopefully!) final review.

geozero/src/mvt/mvt_writer.rs
kylebarron: I still think this would be clearer to not be implemented as `Default`. `Default` should be painfully obvious what it does, and I don't think this meets that bar.

styrowolf:
There seems to be only one blocker for this PR ([link to comment](https://github.com/georust/geozero/pull/264/files/071d1992a329bea7f97a4b47add6b0fc8e4b15f7#r2330378939)):

> Just adding new_unscaled as a new constructor isn't enough in my opinion because the Default impl is still public


What can I do to remove this blocker?

kylebarron:
> > Just adding new_unscaled as a new constructor isn't enough in my opinion because the Default impl is still public
> 
> What can I do to remove this blocker?

My opinion there would be that we should remove the `Default` impl because it's not obvious what it does. And have only `new` and `new_unscaled` constructors.

nyurik (approved):

nyurik:
looks good, thanks!

PR #267, opened 2025-09-15
Cleanup build, add justfile recipes

nyurik:
* Consolidate lint management in the workspace (they will be worked on later)
* Added Cargo.lock - this is a binary, it should contain the lock
* format code a bit
* added justfile with relevant recipes

~- [ ] I added an entry to the project's change log file if knowledge of this change could be valuable to users.
  - Prefix changelog entries for breaking changes with "BREAKING: "~
---


pka (approved):
:+1:

PR #268, opened 2025-10-26
chore(ci): issues in CI hardening

CommanderStorm:
---

This PR fixes a few hardening issues regarding the supply chain CI security of this crate:
- the PR does not have permissions set (=> default is more)
- the shas for CI dependencys are not specified (=> default is that the releases are mutable, so what we consider a version is not actually a version. For shas, it is)
- we are persisting credentials in the checkout step.

I don't think this has effect on consumers (except being more secure), so I did not add an entry to the changelog.

### DIFF
--- a/.github/workflows/audit-check.yml
+++ b/.github/workflows/audit-check.yml
@@ -1,0 +1,55 @@
+name: Security audit
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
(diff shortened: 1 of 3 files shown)


nyurik (approved):

CommanderStorm:
seems to have been merged already

PR #270, opened 2025-10-26
chore(ci): split the linting and testing CI

CommanderStorm:
---

CI is by all standards relatively slow (6m) compared to what it does.
This is because we are doing all checks in one thread on the same worker.

We don't need to => this PR moves all linting tasks into its own job => reducing the total CI time by ~2m.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -86,18 +86,18 @@ jobs:
         env:
           DATABASE_URL: "${{ steps.pg.outputs.connection-uri }}?sslmode=disable"
           
-    # This final step is needed to mark the whole workflow as successful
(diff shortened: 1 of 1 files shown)


CommanderStorm:
clippy and cargo-test are not really the same target to compile for.
I think the results speak for themselves:

image

Are the extra 2m important? Likely not. Still a nice to have

nyurik:
hehe, well, true, except that it is actually 8 min (two instances 4m each... or less?), but the value is small enough and human time is definitely more valuable

nyurik (approved):

CommanderStorm:
Yes, both share the same 1-2m of installing their dependencys.

PR #277, opened 2025-12-18
fix: allow CCW polygons in MVT

keiya01:
---

Thank you for this amazing project.

I encountered an unexpected error when parsing MVT geometry using `ToGeo`.
The error was `MvtError::GeometryFormat`, and it occurred when the MVT geometry contained a counter-clockwise (CCW) exterior polygon.

My proposal may not strictly follow [the MVT specification](https://github.com/mapbox/vector-tile-spec/tree/master/2.1#4344-polygon-geometry-type), but it should handle a wider range of real-world data.


CommanderStorm (approved):
I guess yes, makes sense to be more lenient on input..

PR #281, opened 2026-02-21
ci: add non-blocking Codecov coverage upload job

herostrat:
---
I added a coverage job to CI that generates llvm-cov output and uploads it to Codecov without enforcing a coverage threshold.
For this I introduced a dedicated coverage step in the Linux workflow, enabled continue-on-error behavior, and configured Codecov upload to not fail CI on upload issues.

This results in that we get immediate coverage visibility and trend tracking with zero merge friction, which is a low-risk first step before any future gating policy.


CommanderStorm (approved):
LGTM

PR #282, opened 2026-02-21
test(postgis): replace ignored tests with env-based skip

herostrat:
---
I removed blanket ignore usage for PostGIS integration tests and switched to environment-aware execution.
This is done by adding a shared runtime skip helper that skips only when DATABASE_URL is missing, updated CI to run the PostGIS test target directly, and documented the behavior in the README.
I changed this to keeps local developer workflows smooth while making CI behavior explicit and deterministic, improving test maintainability and reducing hidden test coverage gaps.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -82,7 +82,7 @@ jobs:
         shell: bash
         run: |
           psql -P pager=off -v ON_ERROR_STOP=1 -f geozero/tests/data/postgis.sql $DATABASE_URL
-          cargo test -p geozero --all-features -- --ignored postgis --test-threads 1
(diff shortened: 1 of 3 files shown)


CommanderStorm:
> I changed this to keeps local developer workflows smooth while making CI behavior explicit and deterministic, improving test maintainability and reducing hidden test coverage gaps.

I am not quite convinced that an env var vs `--ignored postgis` is an improvement. I think it an side-grade unless I am missing something.

An cfg would be an upgrade yes, because no spam that tests are ignored, but this? not sure.

CommanderStorm:
if there is a reason you want this, we can reopen

PR #284, opened 2026-03-18
feat: add GeozeroGeometry and FromWkb support for postgis_diesel types

georgeboot:
## Summary

- Implements `GeozeroGeometry` for all `postgis_diesel` types (`Point`, `PointZ`, `PointM`, `PointZM`, `AnyPoint`, `LineString<P>`, `Polygon<P>`, `MultiPoint<P>`, `MultiLineString<P>`, `MultiPolygon<P>`, `GeometryCollection<P>`, `GeometryContainer<P>`), enabling conversion to any geozero-supported format (GeoJSON, WKT, WKB, SVG, geo-types, etc.)
- Adds `PostgisDieselWriter` (`GeomProcessor`) and `FromWkb` for `GeometryContainer<P>`, enabling construction of postgis_diesel types from any geozero source
- Adds `wkb::Decode<T>` / `wkb::Encode<T>` Diesel `FromSql`/`ToSql` impls and `impl_diesel_postgis_decode!` / `impl_diesel_postgis_encode!` macros (mirroring existing postgres/sqlx patterns)
- Fully backwards-compatible: existing `Ewkb<B>` Diesel impls and `sql_types` are preserved

## Motivation

The existing `with-postgis-diesel` feature only provided raw `Ewkb<Vec<u8>>` support — opaque WKB bytes with no typed geometry conversion. This meant downstream users had to write hundreds of lines of manual conversion code to bridge postgis_diesel geometries with other formats.

This PR brings the diesel integration to parity with the existing postgres and sqlx integrations, which already have full `GeozeroGeometry` / `FromWkb` support via `wkb::Encode<T>` / `wkb::Decode<T>`.

## Test plan

- [x] 34 unit tests covering all geometry types, dimensions (XY, XYZ, XYM, XYZM), SRID preservation
- [x] Read direction: all postgis_diesel types → WKT and GeoJSON
- [x] Write direction: EWKB roundtrip and GeoJSON → postgis_diesel for Point, PointZ, PointZM, AnyPoint, LineString, Polygon, MultiPoint, MultiLineString, MultiPolygon, GeometryCollection
- [x] `cargo clippy` clean, `cargo test -p geozero` passes, all existing tests unaffected
- [x] Compiles with all PostGIS feature combinations (`with-postgis-diesel`, `with-postgis-postgres`, `with-postgis-sqlx`)


geozero/src/postgis/postgis_diesel.rs
```diff
@@ -1,8 +1,19 @@
...
+    AnyPoint, GeometryCollection, GeometryContainer, LineString, MultiLineString, MultiPoint,
+    MultiPolygon, Point, PointM, PointT, PointZ, PointZM, Polygon,
+};
+use std::io::{Read, Write as _};
+
+// ---------------------------------------------------------------------------
+// SQL types (kept for backwards compatibility)
+// ---------------------------------------------------------------------------
 
 use crate::postgis::postgis_diesel::sql_types::{Geography, Geometry};
```
CommanderStorm: The reason why nobody has reviewed your code is that it looks very much vibed. Could you remove the vibed comments?

geozero/src/postgis/postgis_diesel.rs
```diff
@@ -20,28 +31,1276 @@ pub mod sql_types {
...
+}
+
+impl_from_wkb!(Point);
+impl_from_wkb!(PointZ);
+impl_from_wkb!(PointM);
+impl_from_wkb!(PointZM);
+impl_from_wkb!(AnyPoint);
+
+#[cfg(test)]
+mod tests {
```
CommanderStorm: I don't think the tests quite test distinct things, or? At least the geojson and wkt paths are very distinct. can you consolidate the tests into distinct chunks of functionality testing the part of code added via the public api?

geozero/src/postgis/postgis_diesel.rs
CommanderStorm, line 1: this file is getting quite large. Can you think of a good split point?

geozero/Cargo.toml
```diff
@@ -75,6 +76,7 @@ prost-build = { workspace = true, optional = true }
 geo.workspace = true
 hex.workspace = true
 kdbush.workspace = true
+diesel = { workspace = true, features = ["postgres"] }
```
CommanderStorm: can you make diesel also be an optional? I don't think it should be required for all users.

CommanderStorm (commented):
Sorry, took a bit of time to review.

CI is red and I am not quite confident that the current code is exactly what we want, it is mostly because this is a lot of code.

georgeboot:
hi @CommanderStorm 
Thanks for your review.

I've refactored the approach a bit and taken out the viby parts. First version was to get something out of the door because I needed it elsewhere. But it's in a much better place now.

geozero/src/postgis/postgis_diesel/tests.rs
CommanderStorm, line 1: can you move this to the rest of our integration tests? https://github.com/georust/geozero/tree/main/geozero/tests

georgeboot: done

geozero/tests/postgis.rs
```diff
@@ -398,6 +398,406 @@ mod postgis_sqlx {
     }
 }
 
+#[cfg(feature = "with-postgis-diesel")]
```
CommanderStorm: Why did you put this in the postgis file and not a separate testcase?
I think a spearate would make sense given the size of this...

georgeboot: i figured its related to postgis and honestly personally aren't really bothered by long modules as long as they share the same topic. but pulled it out now

PR #292, opened 2026-05-22
build: upgrade sqlx from 0.8 to 0.9

markhilb:
---


logankaser:
This would also indirectly fix the crates.io docs build issue. Right now:
```
sqlx (with-postgis-sqlx, with-gpkg)
  → sqlx-mysql / sqlx-postgres / sqlx-core
    → sha2 0.10.x / hmac 0.12.x / rsa 0.9.x / md-5 0.10.x / sha1 / signature
      → digest 0.10.7
        → crypto-common 0.1.x
          → generic-array 0.14.x
```
And specifically, `generic-array 0.14.x` causes the doc build to fail due to the removal of `doc_auto_cfg` which it uses.
`crypto-common` in `0.2`, dropped `generic-array` in favor of `hybrid-array`, and that's what `sqlx` `0.9` now depends on.

Ape:
I can confirm this version is working. Any chance of getting it merged?

CommanderStorm (approved):
Thanks

Ape:
Thanks! Could we get a new release for this?

PR #297, opened 2026-06-17
Bound untrusted `Vec::with_capacity` in the WKB to geo-types path (OOM DoS)

LucaCappelletti94:
The WKB reader takes a 32-bit element count straight from the input and hands it to the active `GeomProcessor` as a size hint, and `GeoWriter` (the `to_geo` processor) turned that hint directly into `Vec::with_capacity(count)`. A crafted blob of a few dozen bytes can declare a count near `u32::MAX`, so a tiny malformed value reserves on the order of tens of gigabytes before any coordinate is read.

Any consumer that runs `Ewkb(untrusted_bytes).to_geo()` on attacker-controlled input is exposed: under a memory or address-space limit (a containerized database backend being the obvious case) the reservation fails and the allocator aborts the process, which is a remotely triggerable crash from a single poisoned geometry value.

The fix caps the speculative reservation rather than trusting the count. A small `bounded_prealloc::<T>(size_hint)` helper returns the lesser of the requested count and however many `T` fit in a fixed 64 KiB byte budget, and every untrusted `Vec::with_capacity` site in `GeoWriter` now goes through it. This is element-size aware so the byte budget holds regardless of what is being reserved, and it preserves the performance intent of pre-allocation for normal geometries while making the worst case bounded. Correctness is unchanged because the containers still grow to fit the real data as it is actually read, so a genuinely large geometry is handled exactly as before, only without the up-front blind reservation. A malformed blob now runs out of input and returns an `Err` instead of reserving gigabytes.

### DIFF
--- a/geozero/src/geo_types/geo_types_writer.rs
+++ b/geozero/src/geo_types/geo_types_writer.rs
@@ -1,4 +1,5 @@
 use crate::error::{GeozeroError, Result};
+use crate::geometry_processor::bounded_vec;
 use crate::{FeatureProcessor, GeomProcessor, PropertyProcessor};
 use geo_types::{
     Coord, Geometry, GeometryCollection, LineString, MultiLineString, MultiPoint, MultiPolygon,
@@ -79,7 +80,7 @@ impl GeomProcessor for GeoWriter {
 
     fn multipoint_begin(&mut self, size: usize, _idx: usize) -> Result<()> {
         debug_assert!(self.coords.is_none());
-        self.coords = Some(Vec::with_capacity(size));
+        self.coords = Some(bounded_vec::<Coord<f64>>(size)?);
         Ok(())
     }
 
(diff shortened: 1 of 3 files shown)


LucaCappelletti94:
I believe that the `cargo audit` errors are pre-existent on main

djc:
@michaelkirk @nyurik would you have a chance to look at this?

nyurik (changes requested):
Thanks, I like the idea, but I think the API is prone to errors: it is too easy to specify one type in the generic param, but generate a different type for the Vec.  For example, this line would work just fine, allocating the wrong size:

```rust
Vec::<u8>::with_capacity(bounded_prealloc::<Coord<f64>>(size))
```

instead, I think a better approach would be something like this:

```rust
pub fn bounded_vec::<T>(capacity: usize) -> Result<Vec<T>, GeozeroError> {
  validate_capacity::<T>(capacity)?;
  Ok(Vec::with_capacity(...))
}

fn validate_capacity::<T>(capacity: usize) Result<(), GeozeroError> {
  todo!("validate")
}
```

the second function can be used for unit testing, but it should not be used by any actual code except the `bounded_vec`.

LucaCappelletti94:
Updated as described

nyurik (approved):

PR #299, opened 2026-07-04
Limit WKB geometry nesting depth to prevent stack overflow

LucaCappelletti94:
The WKB reader recurses once per container nesting level with no depth bound. Because WKB data typically comes from external sources (user uploads, database queries, network requests), an attacker can craft a small binary payload of deeply nested `GeometryCollection`s that overflows the stack and aborts the process. A developer integrating `geozero` to decode user-supplied geometry has no way to know the decoder is vulnerable to stack-overflow denial of service.

`Ewkb`, `Wkb`, `GpkgWkb`, `SpatiaLiteWkb`, and `MySQLWkb` are all affected because they share the same reader. On a default 8 MiB stack, roughly 14 KiB of input is enough to crash the process.

This adds a `WKB_MAX_NESTING_DEPTH` constant (128) and threads a depth counter through `process_wkb_geom_n` and the curve helpers (`process_compoundcurve`, `process_curve`, `process_curvepolygon`). When the limit is exceeded the decoder returns a new `WkbMaxNestingDepth` error instead of recursing further. Real geometries nest only a handful of levels, so the cap never bites legitimate data.

### DIFF
--- a/geozero/src/error.rs
+++ b/geozero/src/error.rs
@@ -40,6 +40,9 @@ pub enum GeozeroError {
     #[error("I/O error `{0}`")]
     IoError(#[from] std::io::Error),
     // Format Specific
+    #[cfg(feature = "with-wkb")]
+    #[error("WKB geometry nesting exceeds maximum depth of {0}")]
+    WkbMaxNestingDepth(u32),
     #[cfg(feature = "with-csv")]
     #[error("CSV error `{0}`")]
     CsvError(#[from] crate::csv::CsvError),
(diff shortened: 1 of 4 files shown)


LucaCappelletti94:
Just to provide the twin note to the other comment, I suggest we decide one standard approach for recursion here https://github.com/georust/wkt/pull/152 and just adapt whatever is chosen.

CommanderStorm (approved):

LucaCappelletti94:
@CommanderStorm did you check out the conversation on recursion here: https://github.com/georust/wkt/pull/152 ?

As I noted in the [previous comment](https://github.com/georust/geozero/pull/299#issuecomment-4916238486), I believe that whatever decision is made for any of these parser crates, we should likely adapt everywhere, and I do not think a consensus was reached at this time in the linked PR.

michaelkirk:
This seems fine to me as is. 

Let's start simple and revisit if someone actually has the use case of parsing super deeply nested GeometryCollection.

Even then, I'd be cautious about unduly complicating the code/dependency tree without some pretty compelling evidence.

PR #301, opened 2026-07-07
chore: cleanup CI, build, justfile

nyurik:
Cleaning up a bit and making clippy pass, update dependencies to avoid any CI security flags.  Codewise there are no changes other than the clippy lints.

Removed MSRV as it mostly gets in the way when we don't have MSRV check and have a lock file. Worst case - it won't compile by an older compiler, which is an identical behavior as if rust-version is set.

---


lnicola:
> Removed MSRV as it mostly gets in the way

How?

> Worst case - it won't compile by an older compiler, which is an identical behavior as if rust-version is set.

It's not identical:

image

README.md
```diff
@@ -266,3 +269,25 @@ Full source code: [kdbush.rs](./geozero/tests/kdbush.rs)
...
+
+## License
+
+Licensed under either of
+
+* Apache License, Version 2.0 ([LICENSE-APACHE](LICENSE-APACHE) or <https://www.apache.org/licenses/LICENSE-2.0>)
+* MIT license ([LICENSE-MIT](LICENSE-MIT) or <https://opensource.org/licenses/MIT>)
+  at your option.
+
+### Contribution
```
lnicola:
```suggestion
### Contributing
```

geozero/CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
   * Breaking: `MvtWriter` does not implement `Default` anymore
   * <https://github.com/georust/geozero/pull/264>
 * Updated to 2024 edition
+* Cleanup justfile, editor config, other minor non-coding aspects
```
lnicola, line 23: Probably not worth mentioning, since it doesn't affect users.

nyurik:
@lnicola I had to do the same thing in #297 to get through the security audit CI issue. I have already approved it, but now I stopped merging per your comment until resolved. 

You are right that it is not identical, but in my understanding, it is similar in the sense that if compiler doesn't support some language feature, it simply won't compile. MSRV makes this a bit more restrictive (e.g. it would compile with 1.84, but we don't want it to).  I don't see that much value in that TBH, unless we also add MSRV testing, in which case we should use something like `cargo minimal-versions check --direct` -- this is something we could add later to this project.  Most of my projects use this method: https://github.com/nyurik/http-content-range/blob/1f033052144ca2ccdb5641999d8e8effb7545f10/justfile#L110-L118

lnicola:
Yeah, I'm all for testing MSRV on CI, but I don't think we should removing it in the meanwhile.

nyurik:
I guess we will have to bump it then - because CI audit wouldn't pass it, and cargo update wouldn't get the later versions without it

lnicola:
Sure, I don't have anything against bumping MSRV.

nyurik:
@lnicola ok, fixed

lnicola (approved):
Please squash the `Cargo.lock` changes

nyurik:
ah yes, forgot that we don't use squash on merge here, done

PR #304, opened 2026-07-08
Make README examples runnable as doctests (#119)

### LINKED ISSUE #119: Make readme examples part of `cargo test`
Rust markdown examples are usually included as part of the `cargo test`, but if they are in README, they require some additional snippet of code. Currently, the examples in the readme are non-compilable (they need a few extra lines hidden with the `#` symbols).

01luyicheng:
Refs #119

Sets up the infrastructure to run README code blocks as doctests and enables the first example.

## Changes
- `geozero/src/lib.rs`: add a module
  `#[cfg(all(doctest, feature = "with-geo", feature = "with-geojson"))]`
  whose doc is `include_str!("../../README.md")`, so non-`ignore` README
  blocks run under `cargo test --doc`.
- `README.md`: convert the "Convert a GeoJSON polygon to geo-types and
  calculate centroid" example from ````rust,ignore```` to ````rust```` and add
  the hidden `use` lines it needs (mirroring
  `geozero/tests/geo_types.rs::centroid`).

## Why only one example now
The other README examples stay `rust,ignore` because they need resources not
available to doctests: the GEOS example needs `with-geos` (not default); the
PostGIS examples need a live DB + `with-wkb`/`with-postgis-*`; the FlatGeobuf
examples need the separate `flatgeobuf` crate + a data file/network; the
PathDrawer needs external `flatgeobuf-gpu` UI types; and the Processing-API
examples reference an outdated `geometry.process(.., GeometryType::..)` API.
Enabling those is left as follow-up.

## Feature gating
`with-geo` + `with-geojson` are both default features, so:
- `cargo test -p geozero` (default) -> example runs
- `cargo test -p geozero --no-default-features` -> module absent, no README doctests
- `cargo test --workspace --all-features` -> example runs
- `cargo doc` / clippy -> `cfg(doctest)` is off, module (and README) not included -> no new warnings

## Verification
- `cargo fmt --all -- --check`
- `cargo clippy --workspace --all-features --bins --examples --tests --benches -- -D warnings`
- `cargo test -p geozero`
- `cargo test -p geozero --no-default-features`
- `cargo test --workspace --all-features`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p geozero --all-features --no-deps`

All pass. The new doctest (`readme_doctests (line 47)`) runs and passes under
default features and `--all-features`, and is correctly absent under
`--no-default-features` and during `cargo doc`.

Note: using `Refs #119` (not `Closes`) since this is an incremental step -
only one of several README examples is enabled. Happy to swap to `Closes #119`
if maintainers prefer to auto-close.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,23 @@ Supported dimensions: X, Y, Z, M, T
 | GeoArrow | ✅ | ✅ | Available via the [geoarrow](https://crates.io/crates/geoarrow) crate. |
 | GeoParquet | ✅ | ✅ | Available via the [geoarrow](https://crates.io/crates/geoarrow) crate. |
 
+## Format conversion overview
+
+|               |                         [`GeozeroGeometry`]                                                                              | Dimensions |                        [`GeozeroDatasource`]                                         | Geometry Conversion |            [`GeomProcessor`]                    |
+|---------------|--------------------------------------------------------------------------------------------------------------------------|------------|--------------------------------------------------------------------------------------|---------------------|-------------------------------------------------|
+| CSV           | [csv::Csv], [csv::CsvString]                                                                                             | XY         | -                                                                                    | [ProcessToCsv]      | [CsvWriter](csv::CsvWriter)                     |
+| GDAL          | `gdal::vector::Geometry`                                                                                                 | XYZ        | -                                                                                    | [ToGdal]            | [GdalWriter](gdal::GdalWriter)                  |
+| geo-types     | `geo_types::Geometry<f64>`                                                                                               | XY         | -                                                                                    | [ToGeo]             | [GeoWriter](geo_types::GeoWriter)               |
+| GeoJSON       | [GeoJson](geojson::GeoJson), [GeoJsonString](geojson::GeoJsonString)                                                     | XYZ        | [GeoJsonReader](geojson::GeoJsonReader), [GeoJson](geojson::GeoJson)                 | [ToJson]            | [GeoJsonWriter](geojson::GeoJsonWriter)         |
+| GeoJSON Lines |                                                                                                                          | XYZ        | [GeoJsonLineReader](geojson::GeoJsonLineReader)                                      |                     | [GeoJsonLineWriter](geojson::GeoJsonLineWriter) |
+| GEOS          | `geos::Geometry`                                                                                                         | XYZ        | -                                                                                    | [ToGeos]            | [GeosWriter](geos::GeosWriter)                  |
+| GPX           |                                                                                                                          | XY         | [GpxReader](gpx::GpxReader)                                                          |                     |                                                 |
+| MVT           | [mvt::tile::Feature]                                                                                                     | XY         | [mvt::tile::Layer]                                                                   | [ToMvt]             | [MvtWriter](mvt::MvtWriter)                     |
+| Shapefile     | -                                                                                                                        | XYZM       | [shp::ShpReader]                                                                     |                     |                                                 |
+| SVG           | -                                                                                                                        | XY         | -                                                                                    | [ToSvg]             | [SvgWriter](svg::SvgWriter)                     |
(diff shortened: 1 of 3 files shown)


nyurik:
I usually just add README as the automatic documentation to the main crate, e.g. by adding 

```rust
#![doc = include_str!("../README.md")]
// or
#![cfg_attr(feature = "default", doc = include_str!("../README.md"))]
```

or similar to the `src/lib.rs`

I feel it may be better to just consolidate all the mod-level documentation to README.md

01luyicheng:
Thanks for the suggestion, @nyurik!

The `#![doc = include_str!("../README.md")]` approach is cleaner in the sense that it consolidates the README as the crate-level docs and runs its examples as doctests for free. The reason I went with the `#[cfg(doctest)]` module instead is that `geozero/src/lib.rs` already has its own crate-level doc comment (the `//!` lines at the top), and adding `#![doc = include_str!("../README.md")]` would either duplicate or replace it on docs.rs.

I am happy to switch to the `include_str!` approach if you prefer — I could either:
1. Remove the existing `//!` crate-level docs in `lib.rs` and let the README take over, or
2. Keep the `#[cfg(doctest)]` module as-is so only the examples are tested without changing the rendered docs.

Which direction would you like? Happy to implement either.

nyurik:
I would prefer to consolidate lib.rs and readme content, moving all of it to readme, and including it in lib.rs - this keeps things simpler

01luyicheng:
@nyurik Thanks for the patient guidance! You're right, consolidating keeps things simpler.

I've refactored as you suggested:
1. **Moved all crate-level documentation from `lib.rs` to `README.md`** — the "Format conversion overview" table (with intra-doc links) now lives in the README, which already contained the description, geometry types, and dimensions.
2. **Added `#![cfg_attr(feature = "default", doc = include_str!("../../README.md"))]` at the top of `lib.rs`** — the README is now the single source of truth for crate-level docs. Used `cfg_attr(feature = "default", ...)` (one of the forms you suggested) so that under `--no-default-features` the README is not included as crate docs, which prevents the centroid doctest from failing to compile (it needs `with-geo` + `with-geojson`).
3. **Removed the `#[cfg(doctest)]` module** — README code blocks now serve as doctests directly via `include_str!`.

Verified locally:
- `cargo test --doc -p geozero` — the centroid example doctest passes ✓
- `cargo test --doc -p geozero --no-default-features` — no README doctests extracted, passes ✓
- `cargo fmt --check` and `cargo check` pass ✓

On docs.rs (`all-features = true`), the README is rendered as the crate-level docs with all intra-doc links resolving.

Changes pushed in commit 661f7c0. Looking forward to your review!

01luyicheng:
Quick note on the CI failure: it's not from the README include_str refactor. It's a pre-existing clippy `needless_return_with_question_mark` lint in `gdal_writer.rs` that a newer clippy version started flagging. Fix is in the latest commit — just dropped the redundant `?` after `return Err(...)` in 3 spots. Doctest refactor itself is working: `cargo test --doc` passes and `--no-default-features` correctly excludes the README doctests.

01luyicheng:
Quick correction on my last comment — the "just dropped the redundant `?`"
fix was wrong. Removing the `?` also removed the implicit
`From<GdalError> for GeozeroError` conversion (derived from `#[from]` on
the `GeozeroError::GdalError` variant), so the crate stopped compiling
under `--all-features` with `error[E0308]: mismatched types: expected
GeozeroError, found GdalError`.

The latest commit (`234e797`) uses `.into()` to perform the conversion
explicitly:

```rust
return Err(GdalError::UnsupportedGeometryType(unsupported_type).into());
```

This restores the original behaviour without re-introducing
`return Err(...)?`. CI is now green across all jobs.

CommanderStorm (approved):
Nice change, thank you

PR #313, opened 2026-07-18
ci: Harden GitHub Actions with zizmor

CommanderStorm:


.github/workflows/ci.yml
```diff
@@ -86,17 +94,37 @@ jobs:
         env:
           DATABASE_URL: "${{ steps.pg.outputs.connection-uri }}?sslmode=disable"
           
+  zizmor:
+    name: zizmor - GitHub Actions static analysis
```
CommanderStorm:
```suggestion
    name: zizmor
```

michaelkirk:
So that I'm understanding: zizmor is a github action that audits your github workflow files?

CommanderStorm:
Yes, a static analysis tool for GitHub Actions.
It is good to have enabled because they keep finding new fun ways to make GHA more secure => enabling in CI is worth it imo.

CommanderStorm:
Here are the docs btw and hardening CI is something that I would want given that supply chain attacks are something that is happening more and more, even though this is not "actually fixing something an accute issue".

https://docs.zizmor.sh/

michaelkirk:
What I’m still wondering, and what is not immediately answered from that docs link:

Is it
- analyzing just the workflow file? ("You should change some configuration in your workflow .yaml"), or
- the code powering the actions? (“The ***@***.*** (just for example) action has an RCE - the upstream code should change”), or
- both?

CommanderStorm:
A mix of both, but mostly the first only.

For example it has this rule, but I don't actually expect that to report https://docs.zizmor.sh/audits/#known-vulnerable-actions

michaelkirk:
ok, sure let's try it.

michaelkirk (approved):

PR #311, opened 2026-07-18
chore: use fast-mvt for MVT processing

nyurik:
- [x] I agree to follow the project's [code of conduct](https://github.com/georust/.github/blob/main/CODE_OF_CONDUCT.md).
- [x] I added an entry to the project's change log file if knowledge of this change could be valuable to users.
  - Usually called `CHANGES.md` or `CHANGELOG.md`
  - Prefix changelog entries for breaking changes with "BREAKING: "
---

This is mostly vibed with some manual reworks at this stage, and may need a few more passes before merging.  My fast-mvt crate at this stage is significantly faster than anything else available, plus it removes the need for protobuf support

PR #315, opened 2026-07-18
Add MLT (MapLibre Tiles) support via mlt-core

CommanderStorm:


geozero/src/mlt/mlt_error.rs
```diff
@@ -0,0 +1,15 @@
...
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum MltError {
+    /// Error from the `mlt-core` decoder or encoder.
+    #[error("MLT codec error: {0}")]
+    Codec(#[from] mlt_core::MltError),
+    /// The intermediate MVT tile failed to encode.
+    #[error("MVT encoding error: {0}")]
+    MvtEncode(#[from] prost::EncodeError),
```
nyurik: prost shouldn't be here. Instead, lets merge #311 first -- this way this PR can then just replace `fast-mvt` with `mlt-core`, and use the re-exported fast-mvt from it.

PR #320, opened 2026-07-25
Bound untrusted Vec reservations in the Shapefile reader (OOM/panic DoS)

binggao1230:
Extends the untrusted-allocation hardening from #297 (`bounded_vec` /
`MAX_PREALLOC_BYTES`) and #299 to the `with-shp` Shapefile reader — the sibling
reader those passes missed, since the budget helpers were gated behind
`with-geo` and `with-shp` doesn't enable it.

The SHP reader fed `i32` count/size fields decoded from untrusted `.shp`/`.shx`
bytes straight to `Vec::with_capacity` / `vec![0; n]`, unbudgeted and without a
sign check. A ~12-byte record suffices:

- multipatch `record_size = -1` → `-1i32 * 2 as usize` = `0xFFFF_FFFF_FFFF_FFFE`
  → `capacity overflow` panic (debug and release);
- polygon/multipoint `num_points = -1` → `attempt to multiply with overflow`
  panic in debug, multi-GB allocation in release;
- `.shx` `file_length` drives `Vec::<ShapeIndex>::with_capacity` the same way.

A second commit covers the same class one step further in: `record_size = 0` is
non-negative but smaller than the 4-byte shape type it contains, so
`record_size - size_of::<i32>()` underflowed.

Fix reuses the existing budget rather than duplicating it: `geometry_processor.rs`
un-gates `MAX_PREALLOC_BYTES` / `bounded_vec` / `validate_capacity` (they have no
`with-geo` deps); `shp_reader.rs` adds `validate_count` to reject negative counts,
uses checked arithmetic for the `* 2` doublings and the multipart size helpers,
and routes every reservation through `bounded_vec`; `shx_reader.rs` validates
`file_length` and bounds the index reservation.

Tests cover each case returning `Err` instead of panicking, plus a guard that the
existing fixtures still parse. `cargo test -p geozero`, `--no-default-features`,
and `--workspace --features with-shp` pass; fmt and clippy clean.

### DIFF
--- a/geozero/src/geometry_processor.rs
+++ b/geozero/src/geometry_processor.rs
@@ -3,12 +3,14 @@ use crate::error::{GeozeroError, Result};
 
 /// Byte budget for speculatively pre-allocating a container sized from untrusted
 /// input, so a tiny malformed blob can't trigger a huge allocation (OOM DoS).
-#[cfg(feature = "with-geo")]
+///
+/// Shared by every reader/writer that reserves a `Vec` from a size hint decoded off
+/// untrusted bytes (the WKB→geo-types path, the MVT writer, and the Shapefile reader).
+/// Kept ungated so a reader can use it without pulling in `with-geo`.
 pub(crate) const MAX_PREALLOC_BYTES: usize = 16 * 1024 * 1024;
 
 /// `Vec<T>` reserved for `capacity` untrusted elements, after
 /// [`validate_capacity`] has ruled out a size hint big enough to be a lie.
-#[cfg(feature = "with-geo")]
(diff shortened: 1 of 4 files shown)


CommanderStorm:
@gaoflow we merged a few other changes, could you merge those in here? I somewhat remember the changes done here from some of the other PRs, so lets make sure that we don't add unrechable double checks 😉

binggao1230:
Rebased onto main. #322/#327 are disjoint from this — no duplicate checks — but the rebase did produce a duplicate `use std::io::Cursor` (a real compile break, fixed). 29/29 shp-reader tests pass, including the NullShape cases from #327.

PR #322, opened 2026-07-26
Stop the shp reader iterators on the first error

gaoflow:
Both `Iterator` impls in `shp/reader.rs` decide `None` from state that no error path updates: `current_pos` only advances over a record that was read in full, and it is compared against a `file_length` taken straight from the untrusted header. So every error arm returns `Some(Err(..))` without advancing or finishing, and a malformed or truncated `.shp` yields the same error forever — no panic, no diagnostic, just 100% CPU. Repro against the fixture in this repo:

```rust
let shp = &std::fs::read("geozero/tests/data/shp/poly.shp")?[..600]; // truncate
let reader = ShpReader::new(Cursor::new(shp))?;                      // header parses fine
reader.iter_geometries(&mut ProcessorSink::new()).count();           // never returns
```

That also hangs `iter_features(..)?.count()`, the one-liner in the `shp` module docs. The CHANGELOG already describes this bug for `iter_features` ("looping forever / erroring at end-of-file"); the guard added then only fires when the header's `file_length` is honest, and `ShapeIterator` never got it.

Fix: make an error terminal in both iterators. For a sequential byte-stream reader whose source has desynchronised there is nothing to resume from, and it makes the `FusedIterator` impls the file already declares mean something. This covers all three error arms of `ShapeRecordIterator::next`, not just the shape one — I measured the dbase arm and the property-processing arm separately (truncated `.dbf`, and a processor that returns `Err` from `property()`) and both flood identically; fixing `ShapeIterator` alone does not stop them.

Plus the untrusted header arithmetic those extents are built from:

- `Header::read_from` rejects a `file_length` below the 100-byte header every shapefile has. A negative value sign-extended to a ~1.8e19 extent; a value below the header size decoded as an empty dataset with no error at all.
- `iter_geometries` doubles `file_length` in `usize` — in `i32` it overflows for a header near `i32::MAX` words, and that method can't report an error.
- the iterator's `record_size` doubling is checked instead of `as usize * 2`, which overflowed on a negative size.

I reused `InvalidShapeRecordSize` rather than growing the public error enum; happy to add a dedicated variant if you'd prefer.

11 tests, in `#[cfg(test)] mod tests` next to the code (`tests/shp-reader.rs` untouched). 9 of them fail on `main`; they use a bounded pull so a regression fails fast instead of hanging CI. Positive controls cover a well-formed file for both iterators. Independent of #320 — disjoint files, merges either order.

### DIFF
--- a/geozero/src/shp/header.rs
+++ b/geozero/src/shp/header.rs
@@ -7,9 +7,9 @@ use crate::shp::Error;
 use crate::shp::point_z::BBoxZ;
 
 pub(crate) const HEADER_SIZE: i32 = 100;
-const FILE_CODE: i32 = 9994;
+pub(crate) const FILE_CODE: i32 = 9994;
 /// Size of reserved bytes in the header, that have do defined use
-const SIZE_OF_SKIP: usize = size_of::<i32>() * 5;
+pub(crate) const SIZE_OF_SKIP: usize = size_of::<i32>() * 5;
 
 /// struct representing the Header of a shapefile
 /// can be retrieved via the reader used to read
@@ -220,9 +220,9 @@ mod tests {
         let mut h = Vec::with_capacity(HEADER_SIZE as usize);
         h.extend_from_slice(&FILE_CODE.to_be_bytes());
         h.extend_from_slice(&[0u8; SIZE_OF_SKIP]);
-        h.extend_from_slice(&(ShapeType::Polygon as i32).to_be_bytes());
+        h.extend_from_slice(&file_length_16_bit.to_be_bytes());
         h.extend_from_slice(&1000i32.to_le_bytes()); // version
-        h.extend_from_slice(&shape_type.to_le_bytes());
+        h.extend_from_slice(&(ShapeType::Polygon as i32).to_le_bytes());
         h.extend_from_slice(&[0u8; 64]); // 8 x f64 bbox
         assert_eq!(h.len(), HEADER_SIZE as usize);
         h
--- a/geozero/src/shp/header.rs
+++ b/geozero/src/shp/header.rs
@@ -81,21 +81,6 @@ impl Header {
     }
 }
 
-/// A well-formed 100-byte main header with a caller-chosen `file_length`, for
-/// tests that need the readers to accept the header and then hit the record loop.
-#[cfg(test)]
-pub(crate) fn raw_header(file_length_16_bit: i32, shape_type: i32) -> Vec<u8> {
-    let mut h = Vec::with_capacity(HEADER_SIZE as usize);
-    h.extend_from_slice(&FILE_CODE.to_be_bytes());
-    h.extend_from_slice(&[0u8; SIZE_OF_SKIP]);
-    h.extend_from_slice(&file_length_16_bit.to_be_bytes());
-    h.extend_from_slice(&1000i32.to_le_bytes()); // version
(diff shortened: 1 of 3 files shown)


geozero/src/shp/reader.rs
```diff
@@ -160,7 +195,12 @@ impl<T: Read + Seek> ShpReader<T> {
             processor,
             source: self.source,
             current_pos: header::HEADER_SIZE as usize,
-            file_length: (self.header.file_length * 2) as usize,
+            // `file_length` is in 16-bit words. Double it in `usize`: doing it in
+            // `i32` overflows for a header claiming close to `i32::MAX` words, and
+            // this method has no way to report that. `Header::read_from` has
+            // already rejected values below the header size, so the cast is sound.
+            file_length: (self.header.file_length as usize).saturating_mul(2),
```
CommanderStorm: does not add actual value imo, should be clear from the code already

```suggestion
            file_length: (self.header.file_length as usize).saturating_mul(2),
```

geozero/src/shp/header.rs
```diff
@@ -208,4 +214,41 @@ mod tests {
...
+        h.extend_from_slice(&1000i32.to_le_bytes()); // version
+        h.extend_from_slice(&shape_type.to_le_bytes());
+        h.extend_from_slice(&[0u8; 64]); // 8 x f64 bbox
+        assert_eq!(h.len(), HEADER_SIZE as usize);
+        h
+    }
+
+    
+    #[test]
+    fn file_length_shorter_than_the_header_is_rejected() {
```
CommanderStorm, line 232:
```suggestion

    #[test]
    fn file_length_shorter_than_the_header_is_rejected() {
```

CommanderStorm (approved):
I did a few minor nits (renaming done -> encountered_error) and cleaned up the code a bit, but otherwise LGTM

geozero/src/shp/reader.rs
binggao1230: Thanks for the cleanup and for applying the suggestion. The refactor had also left the shp tests uncompilable (shape type written into the file-length field, stale two-arg call sites) — fixed in f2010b4, tests pass.

geozero/src/shp/reader.rs
```diff
@@ -254,3 +282,182 @@ impl ShpReader<BufReader<File>> {
 //         }
 //     }
 // }
+
+#[cfg(test)]
+mod tests {
+    use std::io::Cursor;
+
+    use super::*;
+    use crate::shp::header::{HEADER_SIZE, ShapeType};
```
CommanderStorm:
```suggestion
    use crate::shp::header::{HEADER_SIZE, SIZE_OF_SKIP, FILE_CODE, ShapeType};
```

geozero/src/shp/reader.rs
```diff
@@ -254,3 +282,182 @@ impl ShpReader<BufReader<File>> {
 //         }
 //     }
 // }
+
+#[cfg(test)]
+mod tests {
+    use std::io::Cursor;
+
+    use super::*;
+    use crate::shp::header::{HEADER_SIZE, SIZE_OF_SKIP, FILE_CODE, ShapeType};
```
CommanderStorm:
```suggestion
    use crate::shp::header::{FILE_CODE, HEADER_SIZE, SIZE_OF_SKIP, ShapeType};
```

PR #325, opened 2026-08-01
chore: Implement `FusedIterator` and `size_hint` for the shp iterators

CommanderStorm:
Also `impl ExactSizeIterator + FusedIterator` for the iterators that can.
ALso add `size_hint` for the things that need


geozero/src/shp/reader.rs
```diff
@@ -83,6 +95,22 @@ impl<'a, P: FeatureProcessor, T: Read + Seek + 'a> Iterator for ShapeRecordItera
...
         result
     }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        if self.encountered_error {
+            return (0, Some(0));
+        }
+        let (shapes_lower, shapes_upper) = self.shape_iter.size_hint();
+        let num_records = self.dbf_reader.header().num_records as usize;
+        // Reading past the last dbf record yields one final error item
```
michaelkirk: I'm a little confused by this comment - but assume it's related to the `saturating_add(1)`? Can you add a test for this code path?

michaelkirk (changes requested):
Looks mostly good - but I was confused by one part.

geozero/src/shp/reader.rs
CommanderStorm: Yea, was an halucination, I did not yet fully clean up the PR when you reviewed it

PR #327, opened 2026-08-01
Validate NullShape record sizes

gaoflow:
A NullShape body is exactly its four-byte type field, but it was the only shape branch that did not validate the remaining declared record size. An oversized declaration advanced the SHP iterator past later records without producing an error.

Reject nonzero remaining size for NullShape, matching the other shape readers. The regression builds an in-memory NullShape-plus-Point file: the valid two-word NullShape reaches both records, while declarations of 3, 8, 20, and 1000 words return `InvalidShapeRecordSize` and terminate the iterator.


geozero/tests/shp-reader.rs
```diff
@@ -8,6 +8,68 @@ use geozero::shp::ShpReader;
...
+    let reader = ShpReader::new(Cursor::new(null_shape_then_point(2))).unwrap();
+    let mut records = reader.iter_geometries(&mut sink);
+    assert!(records.next().unwrap().is_ok());
+    assert!(records.next().unwrap().is_ok());
+    assert!(records.next().is_none());
+
+    for size_16_bit in [3, 8, 20, 1000] {
+        let mut sink = ProcessorSink::new();
+        let reader = ShpReader::new(Cursor::new(null_shape_then_point(size_16_bit))).unwrap();
+        let mut records = reader.iter_geometries(&mut sink);
```
CommanderStorm: this seems like an 2-in-1 test, please move the second to an `rstest`, you might need to add that as a dev-dependency.

CommanderStorm (approved):
LGTM, but please split the test into two parts.
We can merge afterwards

binggao1230:
Done — split into an rstest (4 cases for the invalid sizes), kept the valid-size smoke test as a plain #[test]. Added rstest 0.24 as a dev-dependency (workspace.dependencies + geozero/Cargo.toml).

CommanderStorm:
Thanks

